PR #2, opened 2016-02-04
Timer observable

davidwdan:


lib/Rx/Observable/TimerObservable.php
```diff
@@ -0,0 +1,68 @@
...
+                    $observer->onNext($counter++);
+                },
+                $this->dueTime,
+                $this->period
+            );
+        } else {
+            $disposable = $scheduler->schedule(
+                function () use (&$counter, $observer) {
+                    $observer->onNext($counter);
+                    $observer->onCompleted();
```
mbonneau: Should complete be scheduled separately?

davidwdan: You would think, but that's the way that RxJS works...

test/Rx/Functional/Observable/TimerObservableTest.php
```diff
@@ -0,0 +1,184 @@
...
+                onNext(500, 0),
+                onNext(600, 1),
+                onNext(700, 2),
+                onNext(800, 3),
+                onNext(900, 4)
+            ],
+            $results->getMessages()
+        );
+    }
+}
```
mbonneau, line 129: Could you add a test that disposes during the subscription period?

davidwdan, line 129: @mbonneau added some more dispose tests

lib/Rx/Observable.php
```diff
@@ -504,6 +505,19 @@ public function skipUntil(ObservableInterface $other)
...
 
     /**
+     * Returns an observable sequence that produces a value after dueTime has elapsed and then after each period.
+     *
+     * @param $dueTime
+     * @param int|null $period
+     * @param SchedulerInterface $scheduler
+     * @return TimerObservable
+     */
+    public static function timer($dueTime, $period = 0, SchedulerInterface $scheduler = null)
```
asm89: Thinking about the API here. What we currently have seems more like `Interval` to me. [1](http://reactivex.io/documentation/operators/interval.html) `Timer` [2](http://reactivex.io/documentation/operators/timer.html) is emitting one item after an interval, though RxJS has a version that seems to be a combination of both (e.g. an interval that starts _after_ the timer has fired).

Let's add two operators, one `Timer` and one `Interval` that behaves like described and then add some helper methods? E.g.

```
timer($dueTime, $scheduler);
interval($period, $scheduler);
intervalStartingAfter($dueTime, $period, $scheduler); // Can be implemented in terms of the other two?
```

lib/Rx/Observable/TimerObservable.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace Rx\Observable;
+
+use Rx\Observable;
+use Rx\ObserverInterface;
+use Rx\SchedulerInterface;
+
+class TimerObservable extends Observable
```
asm89, line 9: As mentioned above, let's split this in a `Timer` and `Interval`. Should simplify the code a tiny bit. E.g. the `$period > 0` check in `subscribe()` basically shows the that this class has two separate behaviors.

lib/Rx/Observable/TimerObservable.php
asm89: I think we do this in other places as well. We can revisit it later. Good question though.

lib/Rx/Observable/TimerObservable.php
asm89, line 9: Oh, we already have `Interval` with the delay. So we should probably just slim this one down then. :) 

https://github.com/ReactiveX/RxPHP/blob/master/lib/Rx/Observable/IntervalObservable.php

davidwdan, line 9: The other option would be:

```
timer($dueTime, $scheduler);
timerPeriodic($dueTime, $period, $scheduler)
```

And split the observable classes likewise.  

We can accomplish `timerPeriodic` with `interval` and `delay`, so it's probably not necessary.  We can always add it later, if we need to.

asm89, line 9: Yeah, let's revisit adding other convenience helpers in a separate issue/PR.

test/Rx/Functional/Observable/TimerObservableTest.php
```diff
@@ -0,0 +1,137 @@
...
+     * @test
+     * @expectedException \InvalidArgumentException
+     */
+    public function timer_one_shot_relative_time_zero_non_int()
+    {
+        $results = $this->scheduler->startWithCreate(function () {
+            return Observable::timer('1', $this->scheduler);
+        });
+
+        $this->assertMessages(
```
asm89: unneeded assert, will never be reached?

PR #3, opened 2016-02-04
Delay was not passing scheduler through

mbonneau:
Delay was not passing scheduler through. This was causing an issue where onCompleted was being called prior to the source subscription completing and before the sourceDisposable was set. onCompleted would attempt to call dispose on a null.

This PR corrects delay and passes the scheduler through, initializes the source disposable in case onCompleted is called prior to subscribe completing. Also has regression tests for both situations.

### DIFF
--- a/test/Rx/Functional/Operator/DelayTest.php
+++ b/test/Rx/Functional/Operator/DelayTest.php
@@ -3,6 +3,7 @@
 namespace Rx\Functional\Operator;
 
 use Rx\Functional\FunctionalTestCase;
+use Rx\Observable;
 
 class DelayTest extends FunctionalTestCase
 {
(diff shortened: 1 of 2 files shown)


test/Rx/Functional/Operator/DelayTest.php
```diff
@@ -270,4 +271,25 @@ public function delay_never()
...
+    }
+
+    /**
+     * @test
+     */
+    public function delay_completes_during_subscribe()
+    {
+        Observable::create(function ($observer, $scheduler) {
+            $observer->onCompleted();
+        })->delay(1, $this->scheduler)->subscribeCallback();
```
asm89: any assertions we can add here besides _not_ crashing?

asm89:
Good catch. Interesting and tricky case!

test/Rx/Functional/Operator/DelayTest.php
mbonneau: I reworked the test cases.

asm89:
:+1: can you squash your commits?

PR #4, opened 2016-02-04
Map updates

davidwdan:
Added index to `map` selector
Added `mapTo` operator
Fixed bug where the `map` operator called `onNext` after it was disposed

### DIFF
--- a/demo/map/mapTo.php
+++ b/demo/map/mapTo.php
@@ -1,0 +1,12 @@
+<?php
+
+require_once __DIR__ . '/../bootstrap.php';
+
+$subscription = Rx\Observable::fromArray([21, 42])
+    ->mapTo(1)
+    ->subscribe($stdoutObserver);
+
+
+//Next value: 1
+//Next value: 1
+//Complete!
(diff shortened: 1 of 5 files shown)


asm89:
Can we add `mapWithIndex()` and have an explicit operator for it? I think there are plenty of cases where you'd want to map without the index, so I'd rather make it explicit for the cases where you want to have the index (in that case it probably makes sense to make the index the first argument?).

`mapWithIndex()` will potentially have problems with `$index` growing over `PHP_INT_MAX`.

davidwdan:
Yeah, that makes sense.

asm89:
@davidwdan Now I think of it, we don't even need the explicit operator, we should just write `mapWithIndex` in terms of `map()`. E.g.

```
public function mapWithIndex(callable $selector)
{
    $index = 0;
    return $this->map(function($value) use ($selector, &$index) {
        return call_user_func_array($selector, [$index++, $value]);
    });
}
```

That should work?

asm89:
:+1:

PR #5, opened 2016-02-05
More GroupByUntil tests with minor fixes to groupByUntil

mbonneau:
I was troubleshooting an issue that I thought was an issue with GroupByUntil. Turned out to be unrelated, but it does bring in a lot more test coverage for that operator, and found one issue.


asm89:
Good catch on the issue and great that you added some extra tests. :+1:

PR #7, opened 2016-02-06
Range Observable scheduler updates

davidwdan:
In the range observable, use the scheduler that's passed through subscribe if one is not passed through the constructor

### DIFF
--- a/lib/Rx/Observable/RangeObservable.php
+++ b/lib/Rx/Observable/RangeObservable.php
@@ -4,6 +4,7 @@ namespace Rx\Observable;
 
 use Rx\Observable;
 use Rx\ObserverInterface;
+use Rx\Scheduler\ImmediateScheduler;
 use Rx\SchedulerInterface;
 
 class RangeObservable extends Observable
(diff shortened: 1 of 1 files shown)


lib/Rx/Observable/RangeObservable.php
```diff
@@ -32,16 +31,23 @@ public function __construct($start, $count, SchedulerInterface $scheduler = null
...
+        $this->scheduler = $scheduler;
 
     }
 
-    public function subscribe(ObserverInterface $observer)
+    public function subscribe(ObserverInterface $observer, SchedulerInterface $scheduler = null)
     {
 
+        if ($this->scheduler !== null) {
+            $scheduler = $this->scheduler;
```
asm89, line 43: Hm, should this actually always prefer the passed in scheduler?

davidwdan, line 43: I guess if we did that, we won't need the scheduler in the constructor, because more often than not a scheduler will be passed in.

asm89, line 43: True, we should revisit it at a later point I guess.

PR #8, opened 2016-02-07
BufferWithCount operator and tests

mbonneau:


lib/Rx/Observable.php
```diff
@@ -839,4 +840,19 @@ public function timeout($timeout, ObservableInterface $timeoutObservable = null,
...
+
+    /**
+     * Projects each element of an observable sequence into zero or more buffers which are produced based on
+     * element count information.
+     *
+     * @param $count
+     * @param int $skip
+     * @return AnonymousObservable
+     */
+    public function bufferWithCount($count, $skip = -1)
```
asm89: Is the `$skip` useful for you? If not I'd propose shipping without it first. It seems to make things more complicated? At least in terms of understanding what the operator does.

If you want to keep it that's fine as well, but can we make `= null` the value for if you don't want to skip?

asm89: Looked at the tests. `$skip` is cool. Let's keep it, but signaling "no skip" with `null` instead of `-1` seems like a good idea to me. What do you think?

mbonneau: `null` makes a lot more sense. I will change that and squash it.

asm89:
:+1:

PR #9, opened 2016-02-07
Updated Readme

davidwdan:


README.md
```diff
@@ -7,9 +7,40 @@ collections and LINQ-style query operators in PHP.
...
+//Completed
+
+```
+
 ## Quick start for demos
 
-Install dependencies using [composer](https://getcomposer.org).
 
 ```bash
 $ composer.phar install
```
asm89, line 46: Can you update this line as well?

Edit: Oh, I thought I read `composer require` instead of `composer.phar require` on top. :)

PR #10, opened 2016-02-07
Update demos which use \Rx\React\Interval

ihor:
Hey guys,

I was running demos to learn how RxPHP works and found that `takeUntil` and `share` demos still use `\Rx\React\Interval` which was removed about one month ago. I replaced it with `\Rx\Observable\IntervalObservable`.

Best regards,
Ihor

### DIFF
--- a/demo/share/share.php
+++ b/demo/share/share.php
@@ -3,11 +3,10 @@
 require_once __DIR__ . '/../bootstrap.php';
 
 $loop = \React\EventLoop\Factory::create();
+$scheduler  = new \Rx\Scheduler\EventLoopScheduler($loop);
 
 //Without Share
-$interval = new \Rx\React\Interval(1000, $loop);
-
-$source = $interval
+$source = \Rx\Observable::interval(1000, $scheduler)
     ->take(2)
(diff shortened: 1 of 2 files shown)


demo/share/share.php
```diff
@@ -3,11 +3,10 @@
 require_once __DIR__ . '/../bootstrap.php';
 
 $loop = \React\EventLoop\Factory::create();
+$scheduler  = new \Rx\Scheduler\EventLoopScheduler($loop);
 
 //Without Share
-$interval = new \Rx\React\Interval(1000, $loop);
-
-$source = $interval
+$source = (new \Rx\Observable\IntervalObservable(1000, $scheduler))
```
davidwdan: I've been trying to use the static methods in the demos (I think that they're easier to read).
Can you change these to `Rx\Observable::interval(1000, $loop)` instead of declaring a new class?

asm89: :+1: `$loop` should be `$scheduler` though :)

ihor: Good point @davidwdan. Changed.

asm89:
:+1:

PR #11, opened 2016-02-07
Retry operator issue

davidwdan:
Check to see if we have a disposable before trying to dispose

### DIFF
--- a/lib/Rx/Operator/RetryOperator.php
+++ b/lib/Rx/Operator/RetryOperator.php
@@ -3,6 +3,7 @@
 namespace Rx\Operator;
 
 use Rx\Disposable\CallbackDisposable;
+use Rx\Disposable\SerialDisposable;
 use Rx\ObservableInterface;
 use Rx\Observer\CallbackObserver;
 use Rx\ObserverInterface;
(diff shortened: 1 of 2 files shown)


lib/Rx/Operator/RetryOperator.php
```diff
@@ -41,7 +41,9 @@ function ($error) use ($observable, $observer, &$disposable, &$getNewObserver, $
                         $observer->onError($error);
                         return;
                     }
-                    $disposable->dispose();
+                    if ($disposable) {
```
asm89: When does this become `null`? `$getNewObserver` is not called until line 56.

davidwdan: Sorry, I missed this comment.  I'll change it to a SerializeDisposable, so it's clear what's going on.

asm89:
@davidwdan Can you add a test that exposes where it previously went wrong?

davidwdan:
@asm89 added a test

lib/Rx/Operator/RetryOperator.php
```diff
@@ -31,18 +32,23 @@ public function __invoke(
         ObservableInterface $observable,
         ObserverInterface $observer,
         SchedulerInterface $scheduler = null
-    ) {
-        $getNewObserver = function () use ($observable, $observer, &$disposable, &$getNewObserver, $scheduler) {
+    )
+    {
```
asm89, line 32: these should be on the same line

lib/Rx/Operator/RetryOperator.php
```diff
@@ -31,18 +32,23 @@ public function __invoke(
...
+
```
asm89: nit: whitespace

PR #12, opened 2016-02-07
Add ability to record and validate output of demo files

asm89:
Over time the RxPHP project has added demos for various observables and
operators. From time to time the demos tend to break because of
refactoring or other changes.

This commit introduces a "test" command inside the demo folder that will
find demos that have an accompanying `.expect` file and ensure that
running the demo yields the same output as is expected. This will ensure
that demos keep working, but it also turns each demo into an end to end
tests for the RxPHP project.

Example usage:

```
$ php demo/test.php
Found 2 demos with an expect file.
[v] demo/combineLatest/combineLatest.php
[v] demo/concat/concat.php
```


asm89:
WIP as @davidwdan is shuffling around the demo code. Once that's done I'll rebase and record the output of the demos.

asm89:
"Proof" that if the contents of the expected file doesn't match the output of the demo it will break CI:

```
$ echo "" > demo/create/create.php.expect
$ php demo/test.php
Found 47 demos with an expect file.
...
[v] demo/count/count.php
[x] demo/create/create.php
[v] demo/defaultIfEmpty/defaultIfEmpty.php
...
$ echo $?
1
```

PR #18, opened 2016-02-09
Added `takeWhile` operator

davidwdan:


lib/Rx/Observable.php
```diff
@@ -366,6 +367,24 @@ public function takeUntil(ObservableInterface $other)
...
 
+
+    /**
+     * Returns elements from an observable sequence as long as a specified condition is true.
+     *
+     * @param callable $predicate - A function to test each source element for a condition.
+     * The callback is called with the following information:
+     *   - the value of the element
+     *   - the index of the element
+     *   - the Observable object being subscribed
```
asm89: Move this to the longer explanation above?

lib/Rx/Observable.php
```diff
@@ -366,6 +367,24 @@ public function takeUntil(ObservableInterface $other)
         });
     }
 
+
+    /**
+     * Returns elements from an observable sequence as long as a specified condition is true.
+     *
+     * @param callable $predicate - A function to test each source element for a condition.
```
asm89: No `-` afaik

lib/Rx/Operator/TakeWhileOperator.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @param ObservableInterface $observable
+     * @param ObserverInterface $observer
+     * @param SchedulerInterface|null $scheduler
+     * @return \Rx\DisposableInterface
+     */
+    public function __invoke(ObservableInterface $observable, ObserverInterface $observer, SchedulerInterface $scheduler = null)
+    {
+        $count = 0;
```
asm89: `$count` can overflow. Should we guard / warn for it?

asm89:
@davidwdan Let's implement this as:
- `takeWhile()` no count
- `takeWhileWithIndex()` with count, perhaps even as the first argument?

What's the use case for passing the source observable as the third argument? None of the tests use it, but I'm interested in the use case in general as well. If there's none at the moment, let's drop it.

test/Rx/Functional/Operator/TakeWhileTest.php
```diff
@@ -0,0 +1,441 @@
...
+            onNext(320, 3),
+            onCompleted(350)
+        ], $results->getMessages());
+
+        $this->assertSubscriptions([
+            subscribe(200, 350)
+        ], $xs->getSubscriptions());
+    }
+
+    private function isPrime($num)
```
asm89, line 408: quite complicated for test code imo :P

PR #20, opened 2016-02-11
retryWhen operator, tests, demo

mbonneau:


lib/Rx/Operator/RetryWhenOperator.php
```diff
@@ -0,0 +1,82 @@
...
+    public function __invoke(
+        ObservableInterface $observable,
+        ObserverInterface $observer,
+        SchedulerInterface $scheduler = null
+    ) {
+        $errors           = new Subject();
+        $when             = call_user_func($this->notificationHandler, $errors->asObservable());
+        $disposable       = new CompositeDisposable();
+        $sourceDisposable = new EmptyDisposable();
+        $innerCompleted   = false;
```
davidwdan, line 38: Is there any reason why these can't be private properties?

test/Rx/Functional/Operator/RetryWhenTest.php
```diff
@@ -0,0 +1,271 @@
...
+
+        $this->assertMessages([
+            onNext(210, 1),
+            onNext(220, 2)
+        ], $results->getMessages());
+
+        $this->assertSubscriptions([
+            subscribe(200, 230)
+        ], $xs->getSubscriptions());
+    }
```
davidwdan, line 270: Should we have some dispose tests?

mbonneau, line 270: I used the tests from the RxJS - I think there could be a few more cases added. I will work on that when I have a chance.

lib/Rx/Operator/RetryWhenOperator.php
asm89, line 38: We shouldn't keep state in the operators themselves I think. Nothing is stopping you from doing:

```
$myOperator = new MyOperator();
$o1->lift($myOperator)->subscribe(/* ... */);
$o2->lift($myOperator)->subscribe(/* ... */);
```

That would get messy?

asm89, line 38: Nevermind. See https://github.com/ReactiveX/RxPHP/pull/26#discussion_r52835458.

mbonneau:
I added a few more tests and caught a few minor issues.

lib/Rx/Operator/RetryWhenOperator.php
```diff
@@ -0,0 +1,104 @@
...
+                [$observer, 'onCompleted']
+            ), $scheduler);
+
+            $disposable->add($sourceDisposable);
+        };
+
+        $whenDisposable = $when->subscribe(new CallbackObserver(
+            function ($x) use ($subscribeToSource, &$sourceError) {
+                if ($sourceError) {
+                    $sourceError = false;
```
asm89, line 88: nit: this already happens in `$subscribeToSource()`?

PR #22, opened 2016-02-11
Added `switchLatest` operator

davidwdan:


asm89:
Implementation lgtm, but I still need to check out the tests.

demo/switch/switchLatest.php.expect
```diff
@@ -0,0 +1,6 @@
+Next value: 0
+Next value: 1
+Next value: 2
+Next value: 3
+Next value: 4
+Complete!
```
davidwdan, line 6: @asm89 I updated the expected output to the correct info.  You can see how it doesn't work with the Immediate Scheduler.  https://travis-ci.org/ReactiveX/RxPHP/jobs/111093463#L236

Once https://github.com/ReactiveX/RxPHP/issues/32 is resolved, this should work.

davidwdan, line 6: I just ran the RxJS example and I'm getting some interesting results.
https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/switch.md

**RxJS 5 output in Node**

```
Next value: 0
Next value: 1
Next value: 2
Next value: 1
Next value: 2
Next value: 3
Next value: 2
Next value: 3
Next value: 4
Complete!
```

**RxJS 5 output in Chrome (correct behavior)**

```
Next value: 0
Next value: 1
Next value: 2
Next value: 3
Next value: 4
Complete!
```

**RxJS 4 output in Node (correct behavior)**

```
Next value: 0
Next value: 1
Next value: 2
Next value: 3
Next value: 4
Complete!
```

The reason for this is RxJS changes the default scheduler based upon the environment, but I don't understand why node would default to immediate.  Immediate blocks, so `switchLatest` behaves exactly like `concatAll` (which is also how it behaves in RxPHP).

I can change the scheduler in the RxPHP demo to use the EventLoop, so that we have a demo that makes sense, but the output is still wrong because of an issue with EventLoopScheduler (@mbonneau has a fix for that)

**RxPHP output with ImmediateScheduler**

```
Next value: 0
Next value: 1
Next value: 2
Next value: 1
Next value: 2
Next value: 3
Next value: 2
Next value: 3
Next value: 4
Complete!
```

**RxPHP output with EventLoopScheduler**

```
Next value: 0
Next value: 2
Next value: 3
Next value: 4
Complete!
```

I'm going to leave it as WIP, until we can get the scheduler issues figured out, so we can have a working demo.

updated: added RxJS v4 output

davidwdan:
All tests are now passing, so this is ready to be reviewed.

PR #25, opened 2016-02-12
[WIP] Static default scheduler

davidwdan:
The purpose of this PR is discuss using a default static scheduler.

usage:

``` php
$scheduler = \Rx\Scheduler::getDefault();  //ImmediateScheduler

//change the default
$scheduler = new EventLoopScheduler($loop);
\Rx\Scheduler::setDefault($scheduler);
```

### DIFF
--- a/lib/Rx/Observable/TimerObservable.php
+++ b/lib/Rx/Observable/TimerObservable.php
@@ -4,6 +4,7 @@ namespace Rx\Observable;
 
 use Rx\Observable;
 use Rx\ObserverInterface;
+use Rx\Scheduler;
 use Rx\SchedulerInterface;
 
 class TimerObservable extends Observable
@@ -25,11 +26,10 @@ class TimerObservable extends Observable
         $this->scheduler = $scheduler;
     }
 
-    public function subscribe(ObserverInterface $observer, SchedulerInterface $scheduler = null)
+    public function subscribe(ObserverInterface $observer)
     {
-        if ($this->scheduler !== null) {
-            $scheduler = $this->scheduler;
-        }
+        $scheduler = $this->scheduler ?: Scheduler::getDefault();
+
         if ($scheduler === null) {
             throw new \Exception("You must use a scheduler that support non-zero delay.");
         }
--- a/lib/Rx/Subject/ReplaySubject.php
+++ b/lib/Rx/Subject/ReplaySubject.php
@@ -6,6 +6,7 @@ use Exception;
 use Rx\Disposable\CallbackDisposable;
 use Rx\Observer\ScheduledObserver;
 use Rx\ObserverInterface;
+use Rx\Scheduler;
 use Rx\Scheduler\ImmediateScheduler;
 use Rx\SchedulerInterface;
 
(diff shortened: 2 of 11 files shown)


lib/Rx/Observable/TimerObservable.php
```diff
@@ -25,11 +26,10 @@ public function __construct($dueTime, SchedulerInterface $scheduler = null)
         $this->scheduler = $scheduler;
     }
 
-    public function subscribe(ObserverInterface $observer, SchedulerInterface $scheduler = null)
+    public function subscribe(ObserverInterface $observer)
```
asm89, line 29: Why did you remove the scheduler here?

lib/Rx/Observable/TimerObservable.php
```diff
@@ -25,11 +26,10 @@ public function __construct($dueTime, SchedulerInterface $scheduler = null)
...
     {
-        if ($this->scheduler !== null) {
-            $scheduler = $this->scheduler;
-        }
+        $scheduler = $this->scheduler ?: Scheduler::getDefault();
```
asm89, line 31: We should probably get rid of `$this->scheduler` here anyway.

asm89:
The idea behind this looks fine to me. Any opinions @mbonneau, @ocramius and @cboden?

Ocramius:
@asm89 this led to incredibly hard-to-debug and coupling in the ZF2 event manager some time ago. Is this just to share the event loop?

asm89:
@Ocramius No, not just to share the event loop, but related I guess. It's mostly a convenience helper so you don't have to pass an explicit scheduler all over the place. That's currently already "supported" but it will then fallback to the `ImmediateScheduler`. That's nice for demos and presentation, but for the real world it doesn't really work well.

Ocramius:
> so you don't have to pass an explicit scheduler all over the place

Makes sense then

lib/Rx/Observable/TimerObservable.php
mbonneau, line 31: I think it is important to have a `$scheduler` parameter on the constructor of time-based operators. If an unsuspecting user subscribes to an observable that uses timing, they may not know to send the appropriate scheduler.

lib/Rx/Subject/ReplaySubject.php
```diff
@@ -6,6 +6,7 @@
 use Rx\Disposable\CallbackDisposable;
 use Rx\Observer\ScheduledObserver;
 use Rx\ObserverInterface;
+use Rx\Scheduler;
 use Rx\Scheduler\ImmediateScheduler;
```
mbonneau, line 10: Unused use statement.

lib/Rx/Observable/TimerObservable.php
cboden, line 29: The scheduler should already be decided by the observable already, not on subscription, I would have thought?

cboden:
My initial reaction reflects @Ocramius'.  It's a compromise between ease of use and "the right way". I'd prefer to see a separate handler for default static stuff than the Scheduler handling its own state AND a global static state.

lib/Rx/Observable/TimerObservable.php
mbonneau, line 29: Right now schedulers "flow" up the chain getting passed from lower operators as they subscribe.

davidwdan:
This is being implemented in the 2.x branch and is superceded by #114

PR #26, opened 2016-02-13
Added `skipWhile` operator

davidwdan:


lib/Rx/Operator/SkipWhileOperator.php
```diff
@@ -0,0 +1,60 @@
...
+    private $index;
+
+    /** @var bool */
+    private $lastResult;
+
+    public function __construct(callable  $predicate)
+    {
+        $this->predicate  = $predicate;
+        $this->index      = 0;
+        $this->lastResult = false;
```
asm89: We should close over the `$index` and `$lastResult` in `__invoke()`. There's nothing stopping a developer from using the same operator instance twice.

asm89:
Can you split into `skipUntil()` and `skipUntilWithIndex()`?

lib/Rx/Operator/SkipWhileOperator.php
asm89: Never mind. Just chatted with @davidwdan on IRC. `lift()` only takes a _factory_ so you always get a new operator. Unless you go full clowntown of course, but then you deserve it? :)

lib/Rx/Operator/SkipWhileOperator.php
```diff
@@ -0,0 +1,56 @@
...
+    public function __invoke(ObservableInterface $observable, ObserverInterface $observer, SchedulerInterface $scheduler = null)
+    {
+        $callbackObserver = new CallbackObserver(
+            function ($value) use ($observer, $observable) {
+                try {
+
+                    if (!$this->lastResult) {
+                        $result = call_user_func_array($this->predicate, [$value, $observable]);
+
+                        $this->lastResult = !$result;
```
asm89: Can we rename this to `$this->isSkipping` or something? E.g.

```
if ($this->isSkipping) {
  $continueSkipping = call_user_func...;
  $this->isSkipping = $continueSkipping;
} else {
  $this->onNext($value);
}
```

demo/skip/skipWhileWithIndex.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+require_once __DIR__ . '/../bootstrap.php';
+
+$observable = Rx\Observable::range(1, 5)
+    ->skipWhileWithIndex(function ($i) {
+        return $i < 3;
+    });
```
mbonneau, line 8: Can we add the value to this example for clarity?

lib/Rx/Observable.php
```diff
@@ -349,6 +350,40 @@ public function skip($count)
     }
 
     /**
+     * Bypasses elements in an observable sequence as long as a specified condition is true and then returns the
+     * remaining elements. The element's index is used in the logic of the predicate function.
```
mbonneau: This is the non-index version - so the index is not used correct?

lib/Rx/Operator/SkipWhileOperator.php
```diff
@@ -0,0 +1,54 @@
...
+
+class SkipWhileOperator implements OperatorInterface
+{
+    /** @var callable */
+    private $predicate;
+
+    /** @var bool */
+    private $isSkipping;
+
+    public function __construct(callable  $predicate)
```
mbonneau: nit: 2 spaces between callabel and $predicate

PR #27, opened 2016-02-13
Add utility to generate documentation for reactivex.io

asm89:
In order to provide up-to-date documentation on reactivex.io we'll
extract the operator documentation directly from the source code. Most
operators already have demo code including expected output in the demo
directory.

In order for an operator to get added to the documentation the docblock
of the method in `Rx\Observable` needs a few additions:
- `@observable` or `@operator` annotation. This will flag the method for
  extraction by the documentation builder.
- `@reactivex` the id of the operator in the reactivex.io repository
- `@demo` (optional) one or multiple annotations containing the path to
  a demo inside the demo directory

For example the `map()` operation:

```
/**
 * It takes a transforming function that operates on each element.
 *
 * @param callable $selector
 * @return AnonymousObservable
 *
 * @demo map/map.php
 * @operator
 * @reactivex map
 */
public function map(callable $selector) { {
    ...
}
```

Usage:

```
$ php docs/build-docs.php
Usage:  <command>
  lint      Verifies that documentation can be loaded
  reactivex Updates the reactivex documentation
```

Note: the command currently assumes a valid checkout of the reactivex
site repository under `docs/reactivex.github.io`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ phpunit.xml
 composer.lock
 composer.phar
 /vendor/
+docs/reactivex.github.io
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,4 @@ before_script: composer install
 script:
   - phpunit
   - php demo/test.php
+  - php docs/build-docs.php lint
--- a/docs/build-docs.php
+++ b/docs/build-docs.php
@@ -1,0 +1,95 @@
+<?php
+
+if (file_exists($file = __DIR__.'/../vendor/autoload.php')) {
+    $autoload = require_once $file;
+} else {
+    throw new RuntimeException('Install dependencies to run doc builder.');
+}
+
+require_once __DIR__ . '/doc-loader.php';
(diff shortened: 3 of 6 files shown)


asm89:
Example of the output after running the script: https://github.com/asm89/reactivex.github.io/commit/e846ee97480601b17b7780829032f246218c04f0.

davidwdan:
Is there anyway that you can generate docs for `@param` and `@return` as well?

Also, I'm getting a lot of these warnings when I run `php docs/build-docs.php`

```
Deprecated: Non-static method Str::firstWord() should not be called statically in /Applications/MAMP/htdocs/Rx.PHP/docs/doc-writer.php on line 43
PHP Deprecated:  Non-static method Str::substringAfter() should not be called statically in /Applications/MAMP/htdocs/Rx.PHP/docs/doc-writer.php on line 43
```

asm89:
@davidwdan Fixed the errors from the script.

How would the `@param` and `@return` values show up in the documentation output? Can we file it as follow up and roll this script? :)

PR #28, opened 2016-02-13
Update `map()` documentation

asm89:
Update the `map()` documentatation to include the relevant annotations
to be included by the documentation builder and add a short description.


asm89:
See https://github.com/ReactiveX/RxPHP/pull/27 for context.

PR #29, opened 2016-02-13
Minimal bufferWithCount demos cleanup

asm89:
Instead of having bigger demo files that demo multiple modes of
operation. Let's keep the demos short. For the expected output folks
should look at the `.expect` file. Also make the stdout observer a bit
smarter by making it json encode all array values before outputting
them.

### DIFF
--- a/demo/bootstrap.php
+++ b/demo/bootstrap.php
@@ -15,9 +15,16 @@ if (file_exists($file = __DIR__.'/../vendor/autoload.php')) {
     throw new RuntimeException('Install dependencies to run test suite.');
 }
 
+function asString($value) {
+    if (is_array($value)) {
+        return json_encode($value);
+    }
+    return (string) $value;
+}
+
 $createStdoutObserver = function ($prefix = '') {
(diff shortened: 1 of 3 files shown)


demo/bootstrap.php
```diff
@@ -15,9 +15,16 @@
     throw new RuntimeException('Install dependencies to run test suite.');
 }
 
+function asString($value) {
+    if (is_array($value)) {
+        return json_encode($value);
```
Ocramius, line 20: seems kinda arbitrary? :|

demo/bootstrap.php
```diff
@@ -15,9 +15,16 @@
...
+    }
+    return (string) $value;
```
Ocramius, line 22: Should probably check with other types of value as well. What about objects?

Ocramius, line 22: EDIT: nvm, saw that these are the demo files

demo/bootstrap.php
asm89, line 20: It is. :+1:

PR #30, opened 2016-02-15
Fix for changing yield behavior in hhvm

robations:
A possible fix. Version comparison has been taken outside the anon function to simplify the IF logic, and avoid re-comparing inside a loop. I've not read any submission guidelines so this is mainly provided as a discussion point!

### DIFF
--- a/lib/Rx/Observable/IteratorObservable.php
+++ b/lib/Rx/Observable/IteratorObservable.php
@@ -26,15 +26,9 @@ class IteratorObservable extends Observable
     {
         $scheduler = $scheduler ?: new ImmediateScheduler();
         $key       = 0;
-
-        return $scheduler->scheduleRecursive(function ($reschedule) use (&$observer, &$key) {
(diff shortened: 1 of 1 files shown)


davidwdan:
Do we know if the behavior in v3.11 is the expected behavior for HHVM, or will it switch back at some point?

robations:
Good question. After searching around I think this is the best clue:

https://github.com/facebook/hhvm/issues/1871

This discusses the issue with `next()` and suggests this was fixed in 3.11 when addressing PHP7 compatibility. The 3v4l.org links at the bottom show the same behaviour carrying through to 3.12.

asm89:
lgtm :+1:

davidwdan:
@robations I'm going to merge this as is.  If we need to refactor it, we can do that at a later date.

robations:
Okay great, I did think of another way to cut this but as you say can be reworked later :)

PR #31, opened 2016-02-19
Max operator, tests, and demos

mbonneau:


test/Rx/Functional/Operator/MaxTest.php
```diff
@@ -0,0 +1,244 @@
...
+            return $xs->max(function () use ($error) {
+                throw $error;
+            });
+        });
+
+        $this->assertMessages([
+            onError(220, $error)
+        ], $results->getMessages());
+    }
+}
```
davidwdan, line 291: Can you add a dispose test?

PR #34, opened 2016-03-06
For Rx\React\Promise::toObservable, wrap non-exceptions on reject.

### LINKED ISSUE #33: Rx\React\Promise::toObservable allows reject to pass non-exceptions to onError
https://github.com/ReactiveX/RxPHP/blob/40edc235ff964e9d035148074a2a48e0eec15e82/lib/Rx/React/Promise.php#L79

There should be something to stop non-exceptions from being passed through to onError.

A behavior for when this happens would need to be worked out - maybe create a Rx\React\RejectedPromiseException that could hold the non-exception value?

mbonneau:
Fixes #33

### DIFF
--- a/lib/Rx/React/Promise.php
+++ b/lib/Rx/React/Promise.php
@@ -76,7 +76,10 @@ final class Promise
                         $subject->onNext($value);
                         $subject->onCompleted();
                     },
-                    [$subject, "onError"]
+                    function ($error) use ($subject) {
+                        $error = $error instanceof \Exception ? $error : new RejectedPromiseException($error);
+                        $subject->onError($error);
+                    }
                 );
 
                 return $subject;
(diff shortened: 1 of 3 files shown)


lib/Rx/React/Promise.php
```diff
@@ -76,7 +76,10 @@ function ($value) use ($subject) {
                         $subject->onNext($value);
                         $subject->onCompleted();
                     },
-                    [$subject, "onError"]
+                    function ($error) use ($subject) {
+                        $error = $error instanceof \Exception ? $error : new RejectedPromiseException($error);
```
davidwdan, line 80: Should we be chaining exceptions, so we're always calling onError with `RejectedPromiseException`?

mbonneau, line 80: I had considered doing that, but I thought it better to not wrap values that are already exceptions as I consider a non-exception value a special case. @asm89 - do you have any thoughts on this?

davidwdan:
LGTM

PR #37, opened 2016-03-10
Events scheduled for the same time at the same time should stay ordered

### LINKED ISSUE #36: EventLoopScheduler Scheduling Bug
Actions that are scheduled from within a scheduled action can be called out of order, which causes unpredictable behavior.

mbonneau:
Fixes #36 

I am not sure if there is a reliable way to test this change. Wanted to get some feedback on this as it seems a little heavy-handed.


davidwdan:
This was taken care of in https://github.com/ReactiveX/RxPHP/issues/50

PR #38, opened 2016-04-06
Added `materialize` and `dematerialize` operators

mbonneau:


test/Rx/Functional/Operator/MaterializeTest.php
```diff
@@ -0,0 +1,202 @@
...
+
+        $results = $this->scheduler->startWithCreate(function () use ($xs) {
+            return $xs->materialize()->dematerialize();
+        });
+
+        $this->assertMessages([
+            onError(250, $error)
+        ], $results->getMessages());
+    }
+}
```
davidwdan, line 248: Can you add some early dispose tests?

PR #41, opened 2016-04-13
Check to see if the interator key is valid instead of null

### LINKED ISSUE #39: IteratorObservable should check Iterator::valid() for completion
https://github.com/ReactiveX/RxPHP/blob/a0ef3c25f7593832cc6b2970ba48a7f15a006ab1/lib/Rx/Observable/IteratorObservable.php#L32

This check does not work in all cases. With SplObjectStorage the key is just an integer that keeps incrementing.

davidwdan:
fixes #39

### DIFF
--- a/test/Rx/Functional/Observable/IteratorObservableTest.php
+++ b/test/Rx/Functional/Observable/IteratorObservableTest.php
@@ -104,6 +104,44 @@ class IteratorObservableTest extends FunctionalTestCase
         ], $results->getMessages());
     }
 
+    /**
+     * @test
+     *
+     * @link https://github.com/ReactiveX/RxPHP/issues/39
+     */
+    public function splObjectStorage_many()
+    {
+
+        if (defined('HHVM_VERSION') && version_compare(HHVM_VERSION, '3.11.0', 'lt')) {
+            $this->markTestSkipped();
+        }
+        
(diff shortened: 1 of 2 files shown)


davidwdan:
I'm not sure what to do about hhvm.  Travis uses v3.6, which is from aug 2015.  All of the tests pass with v3.11+.

davidwdan:
I'm just going to skip the tests that aren't passing with HHVM less than v3.11.0

test/Rx/Functional/Observable/IteratorObservableTest.php
```diff
@@ -104,6 +104,44 @@ public function generator_dispose()
...
 
+    /**
+     * @test
+     *
+     * @link https://github.com/ReactiveX/RxPHP/issues/39
+     */
+    public function splObjectStorage_many()
+    {
+
+        if (defined('HHVM_VERSION') && version_compare(HHVM_VERSION, '3.11.0', 'lt')) {
```
mbonneau, line 115: Does this mean that we only support HHVM at this version or higher? Is there a way to fix the functionality so that it does work in lower versions?

davidwdan: @mbonneau at this point, I don't think it makes any sense to support older versions of hhvm

PR #42, opened 2016-04-16
Fixed issue with combineLatest when using the EventLoop scheduler

davidwdan:
Updated the example to be closer to RxJS


lib/Rx/Operator/CombineLatestOperator.php
```diff
@@ -80,7 +81,7 @@ function ($value) use ($count, &$hasValue, $key, &$values, $observer, &$waitingF
                     }
                 },
                 [$observer, 'onError'],
-                function () use (&$waitingToComplete, $observer) {
+                function () use (&$waitingToComplete, $observer, $key) {
```
mbonneau: $key appears to be unused

PR #44, opened 2016-04-16
Added `timestamp` operator

mbonneau:
This PR adds the `timestamp` operator.

This depends on https://github.com/ReactiveX/RxPHP/pull/43

### DIFF
--- a/demo/timestamp/timestamp.php
+++ b/demo/timestamp/timestamp.php
@@ -1,0 +1,25 @@
+<?php
+
+require_once __DIR__ . '/../bootstrap.php';
+
+$loop = new \React\EventLoop\StreamSelectLoop();
+
+$scheduler = new \Rx\Scheduler\EventLoopScheduler($loop);
+
+$source = \Rx\Observable::interval(1000, $scheduler)
+    ->timestamp()
+    ->map(function ($x) {
+        return $x->value . ':' . $x->timestamp;
+    })
+    ->take(5);
+
+$source->subscribe($createStdoutObserver());
+
+$loop->run();
+
+// Next value: 0:1460781738354
+// Next value: 1:1460781739358
+// Next value: 2:1460781740359
+// Next value: 3:1460781741362
+// Next value: 4:1460781742367
+// Complete!
--- a/demo/timestamp/timestamp.php
+++ b/demo/timestamp/timestamp.php
@@ -8,8 +8,8 @@ $scheduler = new \Rx\Scheduler\EventLoopScheduler($loop);
 
 $source = \Rx\Observable::interval(1000, $scheduler)
     ->timestamp()
-    ->map(function ($x) {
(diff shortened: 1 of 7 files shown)


demo/timestamp/timestamp.php
```diff
@@ -0,0 +1,25 @@
...
+    ->map(function ($x) {
+        return $x->value . ':' . $x->timestamp;
+    })
+    ->take(5);
+
+$source->subscribe($createStdoutObserver());
+
+$loop->run();
+
+// Next value: 0:1460781738354
```
davidwdan, line 20: Can you move the expected output to a .expect file?

lib/Rx/Operator/TimestampOperator.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * TimestampOperator constructor.
+     * @param SchedulerInterface $scheduler
+     */
+    public function __construct(SchedulerInterface $scheduler = null)
+    {
+        $this->scheduler = $scheduler;
+    }
+
```
davidwdan, line 24: nit: extra line

demo/timestamp/timestamp.php
mbonneau, line 20: The expected output will be different every time, that is why there is no expect file.

davidwdan, line 20: Ah, that makes sense

lib/Rx/Operator/TimestampOperator.php
```diff
@@ -0,0 +1,45 @@
...
+            $scheduler = $this->scheduler;
+        }
+
+        if ($scheduler === null) {
+            throw new \Exception("You must use a scheduler with timestamp.");
+        }
+
+        return $observable->subscribe(new CallbackObserver(
+            function ($x) use ($observer, $scheduler) {
+                $observer->onNext((object)["value" => $x, "timestamp" => $scheduler->now()]);
```
davidwdan: Should we use a Timestamped class like [RxJava](https://github.com/ReactiveX/RxJava/blob/fd2da397cfbf73a05f8c8715d18d11633c59d5bb/src/main/java/rx/internal/operators/OperatorTimestamp.java#L55) instead of a stdClass?

demo/timestamp/timestamp.php
```diff
@@ -0,0 +1,25 @@
...
+require_once __DIR__ . '/../bootstrap.php';
+
+$loop = new \React\EventLoop\StreamSelectLoop();
+
+$scheduler = new \Rx\Scheduler\EventLoopScheduler($loop);
+
+$source = \Rx\Observable::interval(1000, $scheduler)
+    ->timestamp()
+    ->map(function ($x) {
+        return $x->value . ':' . $x->timestamp;
```
davidwdan: This needs to be updated to use the `Timestamped` class

PR #48, opened 2016-04-20
Updated annotations and demos for auto generated docs

davidwdan:
Made updates so that we can auto generate docs on http://reactivex.io/. See https://github.com/ReactiveX/RxPHP/pull/27


davidwdan:
You can view what it looks like on this temp site:  http://gemsbok.voryx.net/documentation/operators.html

asm89:
👍 this is really great

docs/doc-writer.php
```diff
@@ -71,10 +71,10 @@ function get_replace_position($lines) {
 }
 
 function build_documentation($docs) {
-    usort(
-        $docs,
-        function($a, $b) { return strcmp($a->methodName, $b->methodName); }
-    );
+//    usort(
```
mbonneau, line 74: Why is this commented out? Is it not needed?

lib/Rx/Observable.php
```diff
@@ -112,54 +118,103 @@ public static function create(callable $subscribeAction)
     /**
      * Returns an observable sequence that produces a value after each period.
```
mbonneau: A little more descriptive would be: "Returns an Observable that emits an infinite sequence of ascending integers starting at 0, with a constant interval of time of your choosing between emissions."

lib/Rx/Observable.php
```diff
@@ -112,54 +118,103 @@ public static function create(callable $subscribeAction)
...
+     * @operator
+     * @reactivex empty-never-throw
      */
     public static function error(\Exception $error)
     {
         return new ErrorObservable($error);
     }
 
     /**
+     * Merges all the observable sequences into a single observable sequence.
```
mbonneau: "Combine an Observable together with another Observable by merging their emissions into a single Observable."

lib/Rx/Observable.php
```diff
@@ -300,6 +390,10 @@ public function select(callable $selector)
      *
      * @param callable $predicate
      * @return \Rx\Observable\AnonymousObservable
+     *
+     * @demo filter/filter.php
+     * @operator
+     * @reactivex filter
      */
     public function filter(callable $predicate)
```
mbonneau, line 398: "Emit only those items from an Observable that pass a predicate test."

lib/Rx/Observable.php
```diff
@@ -534,6 +707,10 @@ public function distinctKey(callable $keySelector, callable $comparer = null)
      *
      * @param callable $comparer
      * @return \Rx\Observable\AnonymousObservable
+     *
+     * @demo distinct/distinctUntilChanged.php
+     * @operator
+     * @reactivex distinct
      */
     public function distinctUntilChanged(callable $comparer = null)
```
mbonneau, line 714: "A variant of distinct that only compares emitted items from the source Observable against their immediate predecessors in order to determine whether or not they are distinct."

lib/Rx/Observable.php
```diff
@@ -606,6 +816,11 @@ public function doOnCompleted(callable $onCompleted)
      * @param $accumulator
      * @param null $seed
      * @return AnonymousObservable
+     *
+     * @demo scan/scan.php
+     * @demo scan/scan-with-seed.php
+     * @operator
+     * @reactivex scan
      */
     public function scan(callable $accumulator, $seed = null)
```
mbonneau, line 823: Remove part about Observable.aggregate.

lib/Rx/Observable.php
```diff
@@ -616,8 +831,13 @@ public function scan(callable $accumulator, $seed = null)
 
     /**
      * Creates an array from an observable sequence.
```
mbonneau: "Creates an observable sequence containing a single element which is an array containing all the elements of the source sequence."

lib/Rx/Observable.php
```diff
@@ -687,6 +923,10 @@ public function asObservable()
      *
      * @param ObservableInterface $observable
      * @return AnonymousObservable
+     *
+     * @demo concat/concat.php
+     * @operator
+     * @reactivex concat
      */
     public function concat(ObservableInterface $observable)
```
mbonneau, line 929: "Concatenate an observable sequence onto the end of the source observable."

mbonneau:
I checked this out - it looks good. I did make a few notes.

docs/doc-writer.php
davidwdan, line 74: It's not needed, because I relying on the order the operators appear in `Observable`

asm89, line 74: Is the output not ok if sorted?

davidwdan, line 74: The issue is that you lose control over what order they appear on the site.

davidwdan, line 74: @asm89 Here's an example of what I'm talking about.  If you take a look at [do](http://gemsbok.voryx.net/documentation/operators/do.html), the alphabetical ordering is `doOnCompleted doOnEach doOnError doOnNext`, which doesn't make sense, it should be `doOnEach doOnNext doOnError doOnCompleted`

asm89, line 74: Check. If you're happy with it now let's ship this then. Down the line we can always add an annotation for the doc ordering. It'd be sad if folks would shuffle around code in `Observable` just because it would suit the docs better.

PR #51, opened 2016-05-09
Reimplement `delay` operator using `materialize` and `timestamp`

mbonneau:
Along with the new EventLoopScheduler, this corrects issues with the delay operator.

### DIFF
--- a/lib/Rx/Operator/DelayOperator.php
+++ b/lib/Rx/Operator/DelayOperator.php
@@ -2,39 +2,48 @@
 
 namespace Rx\Operator;
 
-use Rx\Disposable\CompositeDisposable;
-use Rx\Disposable\EmptyDisposable;
+use Rx\DisposableInterface;
+use Rx\Notification;
+use Rx\Observable\AnonymousObservable;
 use Rx\ObservableInterface;
 use Rx\Observer\CallbackObserver;
 use Rx\ObserverInterface;
 use Rx\SchedulerInterface;
+use Rx\Timestamped;
 
 class DelayOperator implements OperatorInterface
 {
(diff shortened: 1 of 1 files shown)


lib/Rx/Operator/DelayOperator.php
```diff
@@ -2,81 +2,92 @@
...
+                            }
+                            $this->schedulerDisposable = $scheduler->schedule($doScheduledStuff, $this->queue->bottom()->getTimestampMillis() - $scheduler->now());
+                        };
 
-        $disposable->add($sourceDisposable);
+                        $doScheduledStuff();
+                    }
+                },
+                [$observer, 'onError'],
+                function () {
```
davidwdan: You can probably just get rid or the onCompleted argument, since it's replaced with a "noop" in `CallbackObserver`

lib/Rx/Operator/DelayOperator.php
```diff
@@ -2,81 +2,92 @@
...
+        return $observable
+            ->materialize()
+            ->timestamp()
+            ->map(function (Timestamped $x) {
+                return new Timestamped($x->getTimestampMillis() + $this->delayTime, $x->getValue());
+            })
+            ->subscribe(new CallbackObserver(
+                function (Timestamped $x) use ($scheduler, $observer) {
+                    if ($x->getValue() instanceof Notification\OnErrorNotification) {
+                        $x->getValue()->accept($observer);
```
davidwdan, line 64: Should it return after this?

mbonneau, line 64: This is fixed now

PR #53, opened 2016-05-18
Added `flatMapLatest` operator

davidwdan:
RxJs doesn't have any tests, so I didn't add any.  `switchLatest` and `map`'s test should cover all use cases.


mbonneau:
The demo has an expect file which is run as part of the test suite, so I am good with that. If we ever do coverage it will not pick that up though.

PR #55, opened 2016-05-19
Added `race` operator (amb)

davidwdan:


mbonneau:
Why not call this operator `amb`?

cboden:
I'm not sure if this is the direction all of Rx is taking but RxJS is renaming `amb` to `race` in v5 to align it with Observables in ES7.

mbonneau:
`race` seems to make better sense to me - just looked like other Rx implementations were still using `amb` - I am +1 on `race`

PR #56, opened 2016-05-22
Added `takeLast` operator

davidwdan:


lib/Rx/Operator/TakeLastOperator.php
```diff
@@ -0,0 +1,57 @@
...
+
+                if (count($this->items) > $this->count) {
+                    array_shift($this->items);
+                }
+            },
+            [$observer, 'onError'],
+            function () use ($observer) {
+
+                while (count($this->items) > 0) {
+                    $observer->onNext(array_shift($this->items));
```
mbonneau, line 48: Would it be better to schedule this? or use `::fromArray` when `onCompleted` is called?

mbonneau:
Other than the question of scheduling `onCompleted` actions, this LGTM.

lib/Rx/Operator/TakeLastOperator.php
davidwdan, line 48: Not really sure, I just ported over what RxJS has.

PR #58, opened 2016-06-07
Added `repeatWhen` operator

davidwdan:


mbonneau:
@davidwdan Could you add a disposal test?

PR #63, opened 2016-06-16
Added reactiveX min operator

rjhdby:
Added reactiveX min operator

### DIFF
--- a/demo/min/min-with-comparer.php
+++ b/demo/min/min-with-comparer.php
@@ -1,0 +1,18 @@
+<?php
+
+require_once __DIR__ . "/../bootstrap.php";
+
+/* With a comparer */
+$comparer = function ($x, $y) {
+    if ($x > $y) {
+        return 1;
+    } elseif ($x < $y) {
+        return -1;
+    }
+    return 0;
+};
+
+$source = \Rx\Observable::fromArray([1, 3, 5, 7, 9, 2, 4, 6, 8])
+    ->max($comparer);
+
+$subscription = $source->subscribe($createStdoutObserver());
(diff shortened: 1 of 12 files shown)


davidwdan:
Thank you for the PR.  
Just a couple of things before we can review/merge this:
1. Travis is failing because the expected output for `min-with-comparer.php` doesn't match the actual output.
2. Tests for `min` are missing.  You can use [MaxTest.php](https://github.com/ReactiveX/RxPHP/blob/master/test/Rx/Functional/Operator/MaxTest.php) as an example.

rjhdby:
@davidwdan 
I'm see problem here.

Tests failed not on min operator, but on replay operator. My commit have all tests, and they passed.

PHP Warning:  assert(): output does not exist /home/travis/build/ReactiveX/RxPHP/docs/../demo/replay/replay.php.expect failed in /home/travis/build/ReactiveX/RxPHP/docs/doc-loader.php on line 75

Warning: assert(): output does not exist /home/travis/build/ReactiveX/RxPHP/docs/../demo/replay/replay.php.expect failed in /home/travis/build/ReactiveX/RxPHP/docs/doc-loader.php on line 75

demo/min/min-with-comparer.php
```diff
@@ -0,0 +1,18 @@
...
+    if ($x > $y) {
+        return 1;
+    } elseif ($x < $y) {
+        return -1;
+    }
+    return 0;
+};
+
+$source = \Rx\Observable::fromArray([1, 3, 5, 7, 9, 2, 4, 6, 8])
+    ->max($comparer);
```
davidwdan: Should be `min` not `max`.

davidwdan:
The errors you've referenced don't make the tests fail.  They're just there so we know what operators need expect files.

The `min-with-comparer.php` demo expects:

```
Next value: 1
Complete!
```

but the actual is:

```
Next value: 9
Complete!
```

because of https://github.com/ReactiveX/RxPHP/pull/63/files#r67350130

rjhdby:
Yes, my mistake. 

It's magic. Wow! :)

PR #64, opened 2016-06-16
min operator

rjhdby:
added reactiveX min operator

### DIFF
--- a/lib/Rx/Operator/MinOperator.php
+++ b/lib/Rx/Operator/MinOperator.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace Rx\Operator;
+
+use Rx\ObservableInterface;
+use Rx\Observer\CallbackObserver;
+use Rx\ObserverInterface;
+use Rx\SchedulerInterface;
+
+class MinOperator implements OperatorInterface
+{
+    /** @var callable|null */
+    private $comparer;
+
+    /**
+     * MinOperator constructor.
+     * @param $comparer callable
+     */
+    public function __construct(callable $comparer = null)
+    {
+        if ($comparer === null) {
+            $comparer = function ($x, $y) {
+                return $x > $y ? 1 : ($x < $y ? -1 : 0);
+            };
+        }
+
+        $this->comparer = $comparer;
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


davidwdan:
This PR needs functional tests

lib/Rx/Operator/MinOperator.php
```diff
@@ -0,0 +1,71 @@
...
+    public function __invoke(
+        ObservableInterface $observable,
+        ObserverInterface $observer,
+        SchedulerInterface $scheduler = null
+    ) {
+        $previousMin = null;
+        $comparing   = false;
+
+        return $observable->subscribe(new CallbackObserver(
+                                          function ($x) use (&$comparing, &$previousMin, $observer) {
```
davidwdan: Indentation is off.  It should be 12 spaces here.

rjhdby:
@davidwdan What kind of functional tests? Can you explain please?
https://github.com/rjhdby/RxPHP/blob/minOperator/demo/min/min.php
https://github.com/rjhdby/RxPHP/blob/minOperator/demo/min/min-with-comparer.php
Is not enough?

rjhdby:
@davidwdan 
How I understand after read MaxTest, in functional tests first call of onNext is skipped when check results?

```
public function max_number_Some()
    {
        $xs = $this->createHotObservable([
            onNext(150, 1),  <-- If I change it to onNext(150, 5), test still passed normaly
            onNext(210, 3),
            onNext(220, 4),
            onNext(230, 2),
            onCompleted(250)
        ]);

        $results = $this->scheduler->startWithCreate(function () use ($xs) {
            return $xs->max();
        });

        $this->assertMessages([
            onNext(250, 4),      <-- Valid result is 4
            onCompleted(250)
        ], $results->getMessages());
    }
```

Where I can read about right organisation of tests for RxPHP?

davidwdan:
@rjhdby when using the test scheduler, subscriptions don't happen until [200](https://github.com/ReactiveX/RxPHP/blob/master/lib/Rx/Testing/TestScheduler.php#L11), so anything before 200 never gets emitted.

davidwdan:
:+1: LGTM

PR #66, opened 2016-06-18
reactivex sum operator

rjhdby:
reactivex sum operator

### DIFF
--- a/demo/sum/sum_cast.php
+++ b/demo/sum/sum_cast.php
@@ -1,0 +1,13 @@
+<?php
+
+require_once __DIR__ . "/../bootstrap.php";
+
+/*
+ * By specification
+ * true == 1
+ * false == null == 0 
+ */
+$source = \Rx\Observable::fromArray([1, "3", 5e1, true, false, null, 4, 6, 8])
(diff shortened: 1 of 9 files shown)


mbonneau:
I would rather see this (and similar) operators implemented using `reduce` as RxJava does: https://github.com/ReactiveX/RxJavaMath/blob/0.x/src/main/java/rx/math/operators/OperatorSum.java

As far as [dealing with different behavior on unknown values](https://github.com/ReactiveX/RxPHP/pull/66/commits/e3c23136b9e20569eb76d8fab9b2ae32ed614e19#diff-d883cb77e5ee56ec9c137e7288ba4fa5R45), that could just be left to the user - or by allowing a transforming function to be passed in.

Any thoughts @davidwdan @asm89 ?

rjhdby:
@mbonneau I'm thinging about using 'reduce', but decided to not use it.
Argument for reduce - looks good.
Argument against - overheat.

Sum is very simple operator for using additional rx operators inside it, I think.

davidwdan:
@rjhdby I'm closing this because #72 implements this in a much simpler way

PR #69, opened 2016-07-26
Added 'pluck' operator

mbonneau:
This PR adds the `pluck` operator.

There is a difference in the way this version handles objects and arrays without the property or index to be plucked. In RxJS, the operator returns `undefined`, this implementation causes the sequence to error. This is better for PHP as returning any value should be supported and there is not a way to distinguish between a missing property and if it was the value of the property. An additional `map` or `filter` prior to the pluck can give the user control over this circumstance if needed.

Because PHP supports string-indexed arrays, this operator also supports arrays and could be used to pluck specific indexes whether numeric or a string.

### DIFF
--- a/lib/Rx/Observable.php
+++ b/lib/Rx/Observable.php
@@ -1693,4 +1693,29 @@ class Observable implements ObservableInterface
                 return $total / $count;
             }, 0);
     }
+
+    /**
+     * Returns an Observable containing the value of a specified array index (if array) or property (if object) from
+     * all elements in the Observable sequence. If a property can't be resolved the observable will error.
+     *
+     * @param mixed $property
+     * @return Observable
+     *
+     * @demo pluck/pluck.php
+     * @operator
+     * @reactivex pluck
+     */
+    public function pluck($property)
+    {
+        return $this->map(function ($x) use ($property) {
+            if (is_array($x) && isset($x[$property])) {
+                return $x[$property];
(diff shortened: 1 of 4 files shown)


lib/Rx/Observable.php
```diff
@@ -1652,4 +1652,29 @@ public static function race(array $observables)
...
+    /**
+     * Returns an Observable containing the value of a specified array index (if array) or property (if object) from
+     * all elements in the Observable sequence. If a property can't be resolved the observable will error.
+     *
+     * @param mixed $property
+     * @return Observable
+     *
+     * @demo pluck/pluck.php
+     * @operator
+     * @reactivex pluck
```
davidwdan: Pluck is part of [map](http://reactivex.io/documentation/operators/map.html) on reactivex.io not pluck

test/Rx/Functional/Operator/PluckTest.php
```diff
@@ -0,0 +1,162 @@
...
+            onNext(210, 2),
+            onNext(240, 3),
+            onError(290, new \Exception())
+        ], $results->getMessages());
+
+        $this->assertSubscriptions([
+            subscribe(200, 290)
+        ], $xs->getSubscriptions());
+    }
+}
```
davidwdan, line 193: Can you add a test for a numeric index?

PR #80, opened 2016-08-24
Allow subscribe with no arguments

mbonneau:
This PR allows calling `subscribe` on Observables without any arguments.

It is common to start observable sequences without any observer. This allows you to completely leave out arguments.


mbonneau:
While writing some tests I did see that some of the other Observables override `subscribe` and this changes the interface. I am going to close this PR for now.

Issue #81, opened 2016-08-30
doOnEach, doOnNext, doOnError, doOnCompleted do not work as expected


mbonneau:
Currently, the "do" operators do not work as expected after receiving an error or completed.

Reproducing example:

``` php
Observable::just(0)
    ->doOnNext(function ($x) {
        echo "next: " . $x . "\n";
    })
    ->doOnNext(function () {
        echo "completed\n";
    })
    ->repeat(2)
    ->subscribeCallback();
// Output:
// next: 0
// completed
// -------------
// Expected Output:
// next: 0
// completed
// next: 0
// completed
```

This behavior is due to the AbstractObserver class stopping all events once there is a complete or an error.

davidwdan:
Fixed in #82

PR #82, opened 2016-08-30
Create `DoObserver` allowing `do` to be called after error/completion

mbonneau:
This PR addresses #81 by creating a specialized Observer to by used with do.

This is a compromise as there is still some behavior that may not match with what the user expects: If `doOnEach` is used with an Observer that extends `AbstractObserver` such as `CallbackObserver`, they will not receive any events after an error or completion. The other operators (`doOnNext`, `doOnEach`, and `doOnCompleted`) all internally use the DoObserver. This behavior for `doOnEach` has been added to the documentation so that users can hopefully avoid any possible issue.

### DIFF
--- a/lib/Rx/Observable.php
+++ b/lib/Rx/Observable.php
@@ -15,6 +15,7 @@ use Rx\Observable\RangeObservable;
 use Rx\Observable\ReturnObservable;
 use Rx\Observable\TimerObservable;
 use Rx\Observer\CallbackObserver;
+use Rx\Observer\DoObserver;
 use Rx\Operator\AsObservableOperator;
 use Rx\Operator\BufferWithCountOperator;
 use Rx\Operator\CatchErrorOperator;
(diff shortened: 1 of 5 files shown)


lib/Rx/Observable.php
```diff
@@ -805,10 +806,18 @@ public function distinctUntilKeyChanged(callable $keySelector = null, callable $
...
-     *  run arbitrary actions for messages on the pipeline.
+     * Invokes an action for each element in the observable sequence and invokes an action upon graceful
+     * or exceptional termination of the observable sequence.
+     * This method can be used for debugging, logging, etc. of query behavior by intercepting the message stream to
+     * run arbitrary actions for messages on the pipeline.
+     *
+     * When using doOnEach, it is important to note that the Observer may receive additional
+     * events after a stream has completed or errored (such as when useing a repeat or resubscribing).
+     * If you are using an Observable that extends the AbstractObservable, you will not receive these
+     * events. For this special case, use the DoObservable.
```
davidwdan: Should be`DoObserver` not `DoObservable`

lib/Rx/Observable.php
```diff
@@ -805,10 +806,18 @@ public function distinctUntilKeyChanged(callable $keySelector = null, callable $
...
+     *
+     * doOnNext, doOnError, and doOnCompleted uses the DoObservable internally and will receive these
```
davidwdan: Same as above - Should be `DoObserver` not `DoObservable`

PR #84, opened 2016-09-11
Added switchFirst operator

davidwdan:


mbonneau:
There are no tests that switch from the first observable to a subsequent one.

There are no tests for an error on the current subscribed inner observable.

PR #91, opened 2016-10-11
[WIP] Interop Event-Loop Scheduler

davidwdan:
This PR serves as place to discuss using the PHP [Interop Event-Loop](https://github.com/async-interop/event-loop) for the `EventLoopScheduler`, so that RxPHP can be used with [ReactPHP](https://github.com/reactphp/event-loop), [Icicle](https://github.com/icicleio/icicle), [Amphp](https://github.com/amphp/loop), or any project that implements the interop loop.

Usage examples:

With `run` (consistent with the current behavior):

``` PHP
$driver = ReactDriverFactory::createFactoryFromLoop(StreamSelectLoop::class);
Loop::setFactory($driver);

$scheduler = new EventLoopScheduler();
Observable::interval(1000, $scheduler)
    ->take(5)
    ->subscribe($createStdoutObserver());

Loop::get()->run();
```

With `execute`:

``` PHP
$driver = ReactDriverFactory::createFactoryFromLoop(StreamSelectLoop::class);
Loop::setFactory($driver);

Loop::execute(function () use ($createStdoutObserver) {
    $scheduler = new EventLoopScheduler();
    Observable::interval(1000, $scheduler)
        ->take(5)
        ->subscribe($createStdoutObserver());
});
```

Now that we have a static loop, we also need to discuss setting the EventLoopScheduler as the default for operators that require a loop (see https://github.com/ReactiveX/RxPHP/pull/25).  Currently, they default to Immediate and the user needs to manually set it to a supported scheduler, which isn't ideal.

The above example could then be further simplified to:

``` PHP
$driver = ReactDriverFactory::createFactoryFromLoop(StreamSelectLoop::class);
Loop::setFactory($driver);

Observable::interval(1000)
    ->take(5)
    ->subscribe($createStdoutObserver());

Loop::get()->run();
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
         }
     ],
     "require": {
-        "php": "~5.5|~7.0"
+        "php": "~5.5|~7.0",
+        "wyrihaximus/react-async-interop-loop": "dev-master"
     },
     "require-dev": {
         "react/event-loop": "~0.4.1",
--- a/lib/Rx/Scheduler/EventLoopScheduler.php
+++ b/lib/Rx/Scheduler/EventLoopScheduler.php
@@ -2,30 +2,16 @@
 
 namespace Rx\Scheduler;
 
-use React\EventLoop\LoopInterface;
+use Interop\Async\Loop;
 
 final class EventLoopScheduler extends VirtualTimeScheduler
 {
-    /** @var callable */
-    private $timerCallable;
-
     private $nextTimer = PHP_INT_MAX;
-    
+
     private $insideInvoke = false;
 
-    /**
-     * NewEventLoopScheduler constructor.
-     * @param callable|LoopInterface $timerCallableOrLoop
-     */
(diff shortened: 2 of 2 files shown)


davidwdan:
After giving this some more thought I think that these are the steps we need to take:
1. Create a new `InteropLoopScheduler` and leave `EventLoopScheduler` as is, for backwards compatibility.
2. When all of the event loop projects support the interop loop, deprecate `EventLoopScheduler`
3. Use `InteropLoopScheduler` within the new static scheduler as the default async scheduler.

mbonneau:
Just for clarification, RxPHP is already usable by any loop implementation (and can even be used with a simple sleep timer). See: https://github.com/ReactiveX/RxPHP/pull/50#issue-152847397

mbonneau:
This issue is being dealt with in the v2 branch. We are not going to make any changes on v1 for the time being as Interop will be usable as shown in #50 .

PR #92, opened 2016-10-11
Implement `throttle` operator

mbonneau:
In implementing this operator, I changed the behavior from the rxjs4 version.

This operator returns an `Observable` that emits only the first item emitted by the source `Observable` during sequential time windows of a specified duration.

The difference in this implementation is that if items are emitted on the source observable prior to the expiration of the throttle time period, the last item emitted on the source observable will be emitted at the expiration of that period. This guarantees that the most recent item will eventually be emitted.

### DIFF
--- a/lib/Rx/Observable.php
+++ b/lib/Rx/Observable.php
@@ -1783,6 +1783,10 @@ class Observable implements ObservableInterface
      * @param $throttleDuration
      * @param null $scheduler
      * @return AnonymousObservable
+     *
+     * @demo throttle/throttle.php
+     * @operator
+     * @reactivex debounce
      */
     public function throttle($throttleDuration, $scheduler = null)
     {
--- a/lib/Rx/Observable.php
+++ b/lib/Rx/Observable.php
@@ -57,6 +57,7 @@ use Rx\Operator\TakeLastOperator;
 use Rx\Operator\TakeOperator;
 use Rx\Operator\TakeUntilOperator;
 use Rx\Operator\TakeWhileOperator;
+use Rx\Operator\ThrottleOperator;
 use Rx\Operator\TimeoutOperator;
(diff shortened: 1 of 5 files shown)


lib/Rx/Observable.php
```diff
@@ -1771,4 +1772,22 @@ public function pluck($property)
...
+    /**
+     * Returns an Observable that emits only the first item emitted by the source Observable during
+     * sequential time windows of a specified duration.
+     *
+     * If items are emitted on the source observable prior to the expiration of the time period,
+     * the last item emitted on the source observable will be emitted.
+     *
+     * @param $throttleDuration
+     * @param null $scheduler
+     * @return AnonymousObservable
```
davidwdan, line 1785: Can you add the documentation annotations?  It falls into the `debounce` group.

lib/Rx/Observable.php
```diff
@@ -1771,4 +1772,26 @@ public function pluck($property)
             throw new \Exception('Unable to pluck "' . $property . '"');
         });
     }
+
+    /**
+     * Returns an Observable that emits only the first item emitted by the source Observable during
+     * sequential time windows of a specified duration.
+     *
+     * If items are emitted on the source observable prior to the expiration of the time period,
+     * the last item emitted on the source observable will be emitted.
```
davidwdan, line 1781: Should we make a note that our throttle behaves slightly differently than rxjs?

mbonneau: I was not sure about deviating from their implementation in the first place, but I really needed it to behave this way and can't figure out a use case for the other way. It there a correct way?

davidwdan: I guess we have deviated in the past without explicitly noting it and I do think that this is the 'correct' and expected behavior, so disregard.

PR #94, opened 2016-10-13
use now() method in VirtualTimeScheduler::schedule

etienneroudeix:
Will change nothing for VirtualTimeScheduler itself but will fix clock offset issue in EventLoopScheduler

Clock is often out of sync when I use EventLoopScheduler, this patch seems to fix all my troubles.

### DIFF
--- a/lib/Rx/Scheduler/VirtualTimeScheduler.php
+++ b/lib/Rx/Scheduler/VirtualTimeScheduler.php
@@ -32,7 +32,7 @@ class VirtualTimeScheduler implements SchedulerInterface
             return new EmptyDisposable();
         };
 
-        return $this->scheduleAbsoluteWithState($action, $this->clock + $delay, $invokeAction);
+        return $this->scheduleAbsoluteWithState($action, $this->now() + $delay, $invokeAction);
     }
 
     public function scheduleRecursive(callable $action)
--- a/lib/Rx/Scheduler/VirtualTimeScheduler.php
+++ b/lib/Rx/Scheduler/VirtualTimeScheduler.php
@@ -93,7 +93,7 @@ class VirtualTimeScheduler implements SchedulerInterface
 
     public function scheduleRelativeWithState($state, $dueTime, $action)
     {
-        $runAt = $this->clock + $dueTime;
+        $runAt = $this->now() + $dueTime;
 
         return $this->scheduleAbsoluteWithState($state, $runAt, $action);
     }
--- a/test/Rx/Scheduler/EventLoopSchedulerTest.php
+++ b/test/Rx/Scheduler/EventLoopSchedulerTest.php
@@ -109,4 +109,24 @@ class EventLoopSchedulerTest extends TestCase
         
         $this->assertEquals([2], $calls);
     }
+
+    public function testSchedulerWorkedWithScheduledEventOutsideItself()
+    {
+        $loop         = Factory::create();
+        $scheduler    = new EventLoopScheduler($loop);
+
+        $scheduler->start();
+        $start = microtime(true);
+        $called = null;
+
+        $loop->addTimer(0.1, function () use ($scheduler, &$called) {
+            $scheduler->schedule(function () use (&$called) {
+                $called = microtime(true);
+            }, 100);
+        });
+
+        $loop->run();
+
+        $this->assertEquals(0.2, $called-$start, '', 0.02);
+    }
 }


davidwdan:
@etienneroudeix I took a look at this and`now()` returns `$this->time` so I don't see how this change could make any difference.

Can you provide more information on the issue you're experiencing, with code to reproduce the problem?  What is the actual behavior?  What is the expected behavior?

etienneroudeix:
Hello @davidwdan, I'll take time to ease my code and remove my own dependecies to provide an actual buggy sample.

`now()` returns `$this->time` in `VirtualTimeScheduler` but not in `EventLoopScheduler` which get real timestamp from system.

etienneroudeix:
@davidwdan here is an example script.
It just needs `rx/http: ^0.2.0`` library to have a loop
The script also uses demo bootstrap.

``` php
<?php
require_once __DIR__ . '/demo/bootstrap.php';

// need rx/http: ^0.2.0

class EndlessSubject extends \Rx\Subject\Subject
{
    function onCompleted()
    {
    }
}

$loop = \EventLoop\EventLoop::getLoop();

printf("[%s]Start\n", date ('H:i:s'));

$subject = new EndlessSubject();
$subject
    ->lift(function () {
        // the idea here is to do not noify next until we had 10 seconds without any further data
        return function (
            \Rx\ObservableInterface $observable,
            \Rx\ObserverInterface $observer,
            \Rx\SchedulerInterface $scheduler = null
        ) {
            $buffer = [];

            if ($scheduler === null) {
                throw new \Exception("You must use a scheduler that support non-zero delay.");
            }

            $doOnDelay = function () use ($observer, &$buffer) {
                $observer->onNext($buffer);
                $buffer = [];
            };

            $rescheduleDelay = function () use (&$doOnDelayDisposable, $scheduler, $doOnDelay) {
                if (isset($doOnDelayDisposable)) {
                    printf("[%s]Cancel previous scheduler\n", date ('H:i:s'));
                    $doOnDelayDisposable->dispose();
                }

                printf("[%s]Schedule onNext in 10 sec\n", date ('H:i:s'));
                $doOnDelayDisposable = $scheduler->schedule($doOnDelay, 10000);
            };

            return $observable->subscribe(new \Rx\Observer\CallbackObserver(
                function ($x) use ($observer, $rescheduleDelay, &$buffer) {
                    $buffer[] = $x;
                    $rescheduleDelay();
                },
                function ($err) use ($observer, $rescheduleDelay) {
                    $rescheduleDelay();
                    $observer->onError($err);
                },
                function () use ($observer, $rescheduleDelay) {
                    $rescheduleDelay();
                    //$observer->onCompleted();
                }
            ));
        };
    })
    ->subscribeCallback(
        function ($data)  {
            printf("[%s]Got final results\n", date ('H:i:s'));
            print_r($data);
        },
        null,
        null,
        new \Rx\Scheduler\EventLoopScheduler($loop)
    );

$observable = \Rx\Observable::create(function (\Rx\ObserverInterface $observer) use ($loop) {
    $observer->onNext('1');
    $loop->addTimer(1, function () use ($observer) {
        $observer->onNext('1');
    });
    $loop->addTimer(3, function () use ($observer) {
        $observer->onNext('2');
    });
    $loop->addTimer(5, function () use ($observer) {
        $observer->onNext('3');
    });
    $loop->addTimer(9, function () use ($observer) {
        $observer->onNext('4');
        $observer->onCompleted();
    });

})->multicast($subject);

//$observable->subscribe($subject);
$observable->subscribeCallback(
    function ($data) {
        printf("[%s]Got 1 result : %s\n", date ('H:i:s'), $data);
    }
);

$observable->connect();
```

if you run this script with current version you got:

```
[17:30:05]Start
[17:30:05]Schedule onNext in 10 sec
[17:30:05]Got 1 result : 1
[17:30:06]Cancel previous scheduler
[17:30:06]Schedule onNext in 10 sec
[17:30:06]Got 1 result : 1
[17:30:08]Cancel previous scheduler
[17:30:08]Schedule onNext in 10 sec
[17:30:08]Got 1 result : 2
[17:30:10]Cancel previous scheduler
[17:30:10]Schedule onNext in 10 sec
[17:30:10]Got 1 result : 3
[17:30:14]Cancel previous scheduler
[17:30:14]Schedule onNext in 10 sec
[17:30:14]Got 1 result : 4
[17:30:20]Got final results
Array
(
    [0] => 1
    [1] => 1
    [2] => 2
    [3] => 3
    [4] => 4
)
```

The final display did not actually wait for 10 seconds but only 6

Whereas with the fix I submitted :

```
[17:30:40]Start
[17:30:40]Schedule onNext in 10 sec
[17:30:40]Got 1 result : 1
[17:30:41]Cancel previous scheduler
[17:30:41]Schedule onNext in 10 sec
[17:30:41]Got 1 result : 1
[17:30:43]Cancel previous scheduler
[17:30:43]Schedule onNext in 10 sec
[17:30:43]Got 1 result : 2
[17:30:45]Cancel previous scheduler
[17:30:45]Schedule onNext in 10 sec
[17:30:45]Got 1 result : 3
[17:30:49]Cancel previous scheduler
[17:30:49]Schedule onNext in 10 sec
[17:30:49]Got 1 result : 4
[17:30:59]Got final results
Array
(
    [0] => 1
    [1] => 1
    [2] => 2
    [3] => 3
    [4] => 4
)
```

I waited 10 seconds

Thank you for you feedbacks

mbonneau:
This is happening because there is scheduling outside of the event loop scheduler (with `$loop->addTimer(...)`). When the scheduler no longer has events scheduled, it stops running, so the clock does not get updated anymore. The first action scheduled after the scheduler has stopped will be scheduled relative to the last action from before the scheduler was stopped.

Here is a more simplistic example that generates the same issue:

``` php
<?php

require_once __DIR__ . '/vendor/autoload.php';

use React\EventLoop\Factory;
use Rx\Scheduler\EventLoopScheduler;

$loop = Factory::create();

$scheduler = new EventLoopScheduler($loop);

function echoWithTime($x) {
    printf("[%s] $x\n", date ('H:i:s'));
}

$scheduler->start();

echoWithTime(1);

$loop->addTimer(2, function () use ($scheduler) {
    $scheduler->schedule(function () {
        echoWithTime(2);
    }, 2000);
});

$loop->run();
```

This outputs 2 lines that are 2 seconds apart. Intuitively, they should be 4 seconds apart.

The reason `VirtualTimeScheduler` uses `clock` instead of now is to make the entire action execute while the clock is not moving. If this is changed to `now()` then the clock will drift during execution.

This problem is much more likely to cause unexpected results so it makes sense to make this change to `now()`. And for consistency, we should consider the same change in `scheduleRelativeWithState`.

@etienneroudeix - Would you be willing to write a test?

etienneroudeix:
@mbonneau just added a test, I made it with small timer (100ms) to avoid extending test time too much. I had to allow a 0.02 tolerance in assertion otherwise it was failing on your travis php 5.6 environnement (see https://travis-ci.org/ReactiveX/RxPHP/jobs/167594028).
I hope this is the kind of test you expected ?

davidwdan:
LGTM

mbonneau:
@etienneroudeix - This is great - and a good catch.

PR #96, opened 2016-11-06
Refactored `mergeAll` to be consistent with RxJS and `switchLatest`

davidwdan:


mbonneau:
This is a BC break as `mergeAll` would no longer be available as a static.

davidwdan:
@mbonneau  It's more of a bug fix, since the static behavior was wrong and very awkward to use.

mbonneau:
Even though this may be a "fix" - people may still rely on this behavior as this is currently the tested behavior. I would like to find a way to fix this without breaking anything if possible or see it held until a major release.

davidwdan:
I've added support back in for a static `mergeAll` for backwards compatibility.

PR #106, opened 2016-12-28
Added forkJoin operator

martinsik:
This is `forkJoin()` operator implementation based on RxJS 5 (selector function and it's available only as a static method). The tests copy mostly RxJS 4 with exception to selector function which is tested the same way as RxJS 5 tests.


demo/forkJoin-observable/forkJoin-observable.php
```diff
@@ -0,0 +1,14 @@
...
+
+require_once __DIR__ . '/../bootstrap.php';
+
+use Rx\Observable;
+
+$obs1 = Observable::range(1, 4);
+$obs2 = Observable::range(3, 5);
+$obs3 = Observable::fromArray(['a', 'b', 'c']);
+
+$observable = new Rx\Observable\ForkJoinObservable([$obs1, $obs2, $obs3], function($v1, $v2, $v3) {
```
davidwdan: Can you use `Observable::forkJoin` here instead of the class?

lib/Rx/Observable.php
```diff
@@ -234,6 +235,23 @@ public static function mergeAll(ObservableInterface $sources)
...
 
     /**
+     * Runs all observable sequences in parallel and collect their last elements.
+     *
+     * @param array $observables
+     * @param callable|null $resultSelector
+     * @param SchedulerInterface|null $scheduler
+     * @return ForkJoinObservable
+     *
+     * @demo forkJoin-observable/forkJoin-observable.php
```
davidwdan, line 245: I think that just `forkJoin` without -observable is fine

mbonneau:
@martinsik - Do you think you could add a test for throw inside the selector?

lib/Rx/Observable/ForkJoinObservable.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    private $observables;
+
+    private $values = [];
+
+    private $completed = 0;
+
+    private $resultSelector;
+
+    public function __construct(array $observables = [], callable $resultSelector = null) {
```
davidwdan, line 26: $observables is optional here, but required on Observable::forkJoin.  It should probably be required everywhere.

test/Rx/Functional/Observable/ForkJoinObservableTest.php
```diff
@@ -0,0 +1,468 @@
...
+
+        $results = $this->scheduler->startWithCreate(function () use ($xs) {
+            return $xs;
+        });
+
+        $this->assertMessages([
+            onError(220, $error),
+        ], $results->getMessages());
+    }
+}
```
davidwdan, line 526: Can you add some dispose tests?  If you dispose before they all complete, all of the subscription should stop.  ie: https://github.com/ReactiveX/RxPHP/blob/master/test/Rx/Functional/Operator/DelayTest.php#L318

davidwdan (commented):
Just a couple of minor changes.  Overall it looks great!

lib/Rx/Observable/ForkJoinObservable.php
martinsik: I checked the original RxJS 4 and RxJS 5 tests and both test forkJoin with no arguments so it should probably be both optional.

https://github.com/ReactiveX/rxjs/blob/master/spec/observables/forkJoin-spec.ts#L192  
https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/forkjoin.js#L127

davidwdan: Ok, that makes sense

martinsik:
I incorporated everything you've mentioned.

Issue #101, opened 2016-12-18
RxPHP v2.x


davidwdan:
I've started work on RxPHP v2 with the goal of bringing it's behavior closer to RxJS 4/5.  You can see the progress [here](https://github.com/davidwdan/RxPHP/tree/2.x)

Here's a list of the changes I've made so far:

- PHP 7 is a minimum requirement
- It now uses the [interop event-loop](https://github.com/async-interop/event-loop) ( ~~defaults to [ReactPHP](https://github.com/WyriHaximus/reactphp-async-interop-loop), but~~ it'll work with any library that support the interop loop, including ~~[Icicle](https://github.com/icicleio/icicle)~~, [ReactPHP](https://github.com/async-interop/event-loop), [Amphp](https://github.com/amphp/loop) and [KoolKode](https://github.com/koolkode/async))
- It uses a static scheduler (you no longer pass the scheduler through subscribe).  For the most part, the user will not have to worry about the scheduler.
- ~~The event loop auto starts with `register_shutdown_function`.  The less the user has to worry about the underlying event loop and scheduler, the better.~~
- We had some non-standard names for operators, because they were reserved words in PHP, those have all been renamed (`of`, `empty`, `do`, `catch`, `switch`).  Aliases of the old names were left for backwards compatibility.
- Lots of cleanup
- Included [Allow custom operators](https://github.com/ReactiveX/RxPHP/pull/93)

### v1 example
With v1, the user has to know when to create a new scheduler and has to be familiar with event loops, which can be confusing and easy to mess up.

```PHP
$loop      = \React\EventLoop\Factory::create();
$scheduler = new \Rx\Scheduler\EventLoopScheduler($loop);

\Rx\Observable::interval(1000)
    ->take(5)
    ->flatMap(function ($i) {
        return \Rx\Observable::just($i + 1);
    })
    ->subscribe(new \Rx\Observer\CallbackObserver(function ($e) {
        echo $e, PHP_EOL;
    }), $scheduler);

$loop->run();

```

### v2 example
With v2, we pick the correct default scheduler for each operator that requires one, which you can override as needed (setting a scheduler should be the exception, not the rule).  In the example below, `interval` and `of` already have the correct default scheduler, so the user doesn't need to define one.
```PHP
\Rx\Observable::interval(1000)
    ->take(5)
    ->flatMap(function ($i) {
        return \Rx\Observable::of($i + 1);
    })
    ->subscribe(function ($e) {
[...]

PR #110, opened 2017-01-03
[WIP]2.x

davidwdan:
ref: #101


demo/partition/partition.php
```diff
@@ -2,7 +2,7 @@
 
 require_once __DIR__ . '/../bootstrap.php';
 
-list($evens, $odds) = \Rx\Observable::range(0, 10)
+list($evens, $odds) = \Rx\Observable::range(0, 10, new \Rx\Scheduler\ImmediateScheduler())
```
mbonneau: It is not obvious why the ImmediateScheduler is passed in here.

demo/return-observable/return-observable.php
```diff
@@ -2,5 +2,5 @@
 
 require_once __DIR__ . '/../bootstrap.php';
 
-$observable = new Rx\Observable\ReturnObservable(42);
+$observable = new Rx\Observable\ReturnObservable(42, \Rx\Scheduler::getDefault());
```
mbonneau: Why is the scheduler passed in here?

src/Observable.php
```diff
@@ -76,7 +79,34 @@ class Observable implements ObservableInterface
...
+        $observer = new CallbackObserver($onNextOrObserver, $onError, $onCompleted);
+
+        return $this->_subscribe($observer);
+    }
+
+    /**
+     * @param ObserverInterface $observer
+     * @return DisposableInterface
+     */
+    protected function _subscribe(ObserverInterface $observer): DisposableInterface
```
mbonneau: This is never called because the subclasses always override it - should we consider making this abstract?

Also - should we consider renaming this to `subscribeCore`?

src/Observable.php
```diff
@@ -87,9 +117,27 @@ public function subscribe(ObserverInterface $observer, $scheduler = null)
...
+    }
+
+    /**
      * @internal
+     *
+     * @param ObserverInterface $observer
+     * @return bool
      */
-    public function removeObserver(ObserverInterface $observer)
+    public function removeObserver(ObserverInterface $observer): bool
```
mbonneau: This is marked `@internal` and is public. Should this be changed to private or protected?

src/Observable.php
```diff
@@ -135,38 +176,63 @@ public static function create(callable $subscribeAction)
...
      * @param mixed $value Single element in the resulting observable sequence.
-     * @return \Rx\Observable\AnonymousObservable An observable sequence with the single element.
+     * @param SchedulerInterface $scheduler
+     * @return ReturnObservable An observable sequence with the single element.
      *
-     * @demo just/just.php
+     * @demo of/of.php
      * @operator
-     * @reactivex just
+     * @reactivex of
```
mbonneau: Should this still be `just` so it stays in the same docs section on reactivex.io?

src/Observable.php
```diff
@@ -178,24 +244,25 @@ public static function emptyObservable()
...
+     * @param \Exception $error
+     * @param SchedulerInterface $scheduler
      * @return ErrorObservable The observable sequence that terminates exceptionally with the specified exception object.
      *
      * @demo error-observable/error-observable.php
      * @operator
      * @reactivex empty-never-throw
      */
-    public static function error(\Exception $error)
+    public static function error(\Exception $error, SchedulerInterface $scheduler = null): ErrorObservable
```
mbonneau: Should this be `Throwable`? Consider a `try..catch` where you construct an `ErrorObservable` from a caught `Throwable`. I am not 100% up to speed on all the new php 7 stuff - just a thought.

src/Observable.php
```diff
@@ -285,17 +351,19 @@ public static function fromIterator(\Iterator $iterator)
...
      */
-    public static function defer(callable $factory)
+    public static function defer(callable $factory, SchedulerInterface $scheduler = null): AnonymousObservable
     {
-        return (new EmptyObservable())->lift(function () use ($factory) {
-            return new DeferOperator($factory);
-        });
+        return static::empty($scheduler)
+            ->lift(function () use ($factory) {
+                return new DeferOperator(new ReturnsObservable($factory));
```
mbonneau: `ReturnsObservable` is probably not a great name because it is very close to `ReturnObservable` - perhaps `CallableWrapperForSelectorThatReturnsObservable`

src/Observable.php
```diff
@@ -304,15 +372,17 @@ public static function defer(callable $factory)
...
+     * @param SchedulerInterface $scheduler
      * @return RangeObservable
+     * @throws \InvalidArgumentException
      *
      * @demo range/range.php
      * @operator
      * @reactivex range
      */
-    public static function range($start, $count, SchedulerInterface $scheduler = null)
+    public static function range(int $start, int $count, SchedulerInterface $scheduler = null): RangeObservable
```
mbonneau, line 352: Do we need to force `int` here? What if I wanted to do `->range('a', 5)` to get the first 5 letters?

src/Observable.php
```diff
@@ -408,7 +478,7 @@ public function mapTo($value)
      * @operator
      * @reactivex map
      */
-    public function select(callable $selector)
+    public function select(callable $selector): AnonymousObservable
```
mbonneau, line 450: Should the return type of operators be `AnonymousObservable` or just `Observable`. Isn't the use of `AnonymousObservable` an internal implementation detail?

src/Observable.php
```diff
@@ -526,7 +596,7 @@ public function flatMapLatest(callable $selector, SchedulerInterface $scheduler
      * @operator
      * @reactivex skip
      */
-    public function skip($count)
+    public function skip($count): AnonymousObservable
```
mbonneau: `int $count`?

src/Observable.php
```diff
@@ -739,16 +809,16 @@ public function lift(callable $operatorFactory)
...
      * @param $name
      * @param $arguments
-     * @return Observable
-     * 
+     * @return AnonymousObservable
+     *
      * @demo custom-operator/rot13.php
      */
-    public function __call($name, array $arguments)
+    public function __call($name, $arguments): AnonymousObservable
```
mbonneau: Does `array` not belong on this method?

src/Observable.php
```diff
@@ -871,29 +941,40 @@ public function distinctUntilKeyChanged(callable $keySelector = null, callable $
...
      * @return \Rx\Observable\AnonymousObservable
      *
-     * @demo do/doOnEach.php
+     * @demo do/do.php
      * @operator
      * @reactivex do
      *
      */
-    public function doOnEach(ObserverInterface $observer)
+    public function do(ObserverInterface $observer): AnonymousObservable
```
mbonneau: Should `do` allow a `callable` or an `ObservableInterface` the same way that `subscribe` do in the 2.x branch?

src/Observable.php
```diff
@@ -1381,7 +1461,7 @@ public function retry($retryCount = -1)
      * @operator
      * @reactivex retry
      */
-    public function retryWhen(callable $notifier)
+    public function retryWhen(callable $notifier): AnonymousObservable
     {
         return $this->lift(function () use ($notifier) {
             return new RetryWhenOperator($notifier);
```
mbonneau: Should this use `ReturnsObservable` on the notifier?

src/Observable.php
```diff
@@ -1473,7 +1553,7 @@ public function repeatWhen(callable $notifier)
      * @param SchedulerInterface $scheduler
      * @return AnonymousObservable
      */
-    public function subscribeOn(SchedulerInterface $scheduler)
+    public function subscribeOn(SchedulerInterface $scheduler): AnonymousObservable
```
mbonneau, line 1529: I question the usefulness of `subscribeOn`. This operator is intended more for multi-threaded environments. How does using this change the behavior? And how does this relate to `observeOn`?

src/Observable.php
```diff
@@ -1522,13 +1602,14 @@ public function timeout($timeout, ObservableInterface $timeoutObservable = null,
...
      * @return AnonymousObservable
+     * @throws \InvalidArgumentException
      *
      * @demo bufferWithCount/bufferWithCount.php
      * @demo bufferWithCount/bufferWithCountAndSkip.php
      * @operator
      * @reactivex buffer
      */
-    public function bufferWithCount($count, $skip = null)
+    public function bufferWithCount(int $count, $skip = null): AnonymousObservable
```
mbonneau: `int $skip`?

src/Observable.php
```diff
@@ -1541,49 +1622,62 @@ public function bufferWithCount($count, $skip = null)
...
-     * @demo catch/catchError.php
+     * @demo catch/catch.php
      * @operator
      * @reactivex catch
      */
-    public function catchError(callable $selector)
+    public function catch (callable $selector): AnonymousObservable
     {
         return $this->lift(function () use ($selector) {
             return new CatchErrorOperator($selector);
```
mbonneau: `ReturnsObservable`?

src/Observable/GroupedObservable.php
```diff
@@ -22,11 +21,11 @@ public function __construct($key, ObservableInterface $underlyingObservable, Dis
             $this->underlyingObservable = $underlyingObservable;
         } else {
             $this->underlyingObservable = new AnonymousObservable(
-                function ($observer, $scheduler) use ($mergedDisposable, $underlyingObservable) {
+                function ($observer) use ($mergedDisposable, $underlyingObservable) {
                     // todo, typehint $mergedDisposable?
```
mbonneau: This comment should be removed. There are also a couple other todo lines that need to be removed.

src/Operator/MaxOperator.php
```diff
@@ -64,8 +53,8 @@ function () use (&$comparing, &$previousMax, $observer) {
                     return;
                 }
 
-                $observer->onError(new \Exception("Empty"));
+                $observer->onError(new \Exception('Empty'));
```
mbonneau: Could we use a better exception or at least a better message here?

src/Operator/MinOperator.php
```diff
@@ -64,8 +53,8 @@ function () use (&$comparing, &$previousMin, $observer) {
                     return;
                 }
 
-                $observer->onError(new \Exception("Empty"));
+                $observer->onError(new \Exception('Empty'));
```
mbonneau: Same.

src/Operator/TimeoutOperator.php
```diff
@@ -4,54 +4,37 @@
...
-    public function __construct($timeout, ObservableInterface $timeoutObservable = null,  SchedulerInterface $scheduler = null)
+    public function __construct(int $timeout, ObservableInterface $timeoutObservable = null, SchedulerInterface $scheduler)
     {
         $this->timeout           = $timeout;
         $this->scheduler         = $scheduler;
         $this->timeoutObservable = $timeoutObservable;
 
         if ($this->timeoutObservable === null) {
-            $this->timeoutObservable = new ErrorObservable(new \Exception("timeout"));
+            $this->timeoutObservable = new ErrorObservable(new \Exception('timeout'), $scheduler);
```
mbonneau, line 29: Maybe make a `TimeoutException` so downstream can catch that specific case.

src/ReturnsObservable.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Rx;
+
+class ReturnsObservable
```
mbonneau: Name issue above.

src/Scheduler.php
```diff
@@ -0,0 +1,53 @@
...
+    public static function getDefault(): SchedulerInterface
+    {
+        if (!static::$default) {
+            static::$default = new EventLoopScheduler();
+        }
+
+        return static::$default;
+    }
+
+    public static function setDefault(SchedulerInterface $scheduler)
```
mbonneau, line 23: The static schedulers should not be settable. If they are - they should only be set once or use a factory.

src/Scheduler/ImmediateScheduler.php
```diff
@@ -48,20 +49,18 @@ public function scheduleRecursive(callable $action)
 
     /**
      * @inheritDoc
+     * @throws \Exception
      */
-    public function schedulePeriodic(callable $action, $delay, $period)
+    public function schedulePeriodic(callable $action, $delay, $period): DisposableInterface
     {
-        throw new \Exception("ImmediateScheduler does not support a non-zero delay.");
+        throw new \Exception('ImmediateScheduler does not support a non-zero delay.');
```
mbonneau, line 56: These should be `InvalidArgumentExceptions`

src/bootstrap.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+use Interop\Async\Loop;
+
+register_shutdown_function(function () {
```
mbonneau, line 5: I would like to discuss auto-starting the scheduler for everything.

src/bootstrap.php
```diff
@@ -0,0 +1,7 @@
...
+    Loop::execute(function () {}, Loop::get());
```
mbonneau, line 6: Is there no `Loop::run()`?

test/Rx/Functional/Observable/IteratorObservableTest.php
```diff
@@ -15,7 +14,7 @@ public function it_schedules_all_elements_from_the_generator()
     {
         $generator = $this->genOneToThree();
 
-        $xs = new \Rx\Observable\IteratorObservable($generator);
+        $xs = new IteratorObservable($generator, $this->scheduler);
```
mbonneau: `Observable::fromIterator`

test/Rx/Functional/Observable/ReturnObservableTest.php
```diff
@@ -5,25 +5,26 @@
...
 use Rx\Observable\ReturnObservable;
 use Rx\Observer\CallbackObserver;
+use Rx\Scheduler;
 
 class ReturnObservableTest extends FunctionalTestCase
 {
     public function testReturnObservableSubscribeTwice()
     {
-        $o = new ReturnObservable("The Value");
+        $o = new ReturnObservable('The Value', Scheduler::getImmediate());
```
mbonneau, line 14: `Observable::of`

mbonneau (commented):
This looks good - most of the comments I made were on things that were not even changed, just things that we may want to change since we are going to a new major version.

mbonneau:
@davidwdan - Because this is a big PR, I think we should close this and make all the issues smaller PRs against the 2.x branch that way the PRs don't take a huge time commitment to review.

When it comes time to release, master can be moved to this branch and 1.x tagged for maintenance.

src/Observable.php
davidwdan: I think that it's only used for testing.

src/Observable.php
davidwdan: Yes, it should be 'just'

src/Observable.php
davidwdan: Can we think of a longer name?  That one is too short

src/Observable.php
davidwdan: Good point.  same applies for the RangeObservable class

src/Observable.php
davidwdan: That's a good idea

src/Observable.php
davidwdan: yes

src/Observable.php
davidwdan: We don't have observeOn.  Now that the scheduler isn't being passed through subscribe, I don't know if it does anything useful.

src/bootstrap.php
davidwdan: Can you open an issue?

src/bootstrap.php
davidwdan: There is, but from my understanding, this is the right way to do this.

davidwdan:
@mbonneau can you take care of the exceptions and I'll hit the other items

demo/return-observable/return-observable.php
davidwdan: The scheduler is no longer optional for Observables and Operators that require a scheduler.

src/Observable.php
davidwdan: We also need some tests for alpha ranges

src/Observable.php
mbonneau: We should get rid of it in v2 - it is easier to add back than it is to get rid of it later.

demo/return-observable/return-observable.php
mbonneau: Should we encourage `Observable::of` instead?

src/Observable.php
mbonneau: So it shouldn't be public right?

davidwdan: It probably only belongs on Subjects anyway, since an observable will only ever have one observer

src/Observable.php
mbonneau: `ObservableFactory` or `ObservableFactoryWrapper`

test/Rx/Functional/Observable/ReturnObservableTest.php
davidwdan: That is the ReturnObservable test, so it's fine the way it is.
We do need to add some Observable::of tests in Operator

src/Observable.php
davidwdan: I think AnonymousObservable makes sense, since that's what's being returned

mbonneau: I think changing this to `Observable` makes more sense. It would introduce a break in the API if the internal details of how the operator works are changed. (probably less likely on this particular operator)

Returning the more restrictive `AnonymousObservable` does not gain you anything.

src/ObservableFactoryWrapper.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Rx;
+
+class ObservableFactoryWrapper
```
mbonneau: final?

src/Observable.php
mbonneau: After looking at the implementation - I think we should leave this as `int` - See below

src/Observable.php
mbonneau: I have moved this to #111

src/Observable.php
mbonneau: This has been moved to #112

src/Operator/TimeoutOperator.php
mbonneau: Moved to #113

src/Scheduler.php
mbonneau: Moved to #114

src/Scheduler/ImmediateScheduler.php
mbonneau: On second thought - I don't think there is much to be gained by adding the specific exception here.

src/bootstrap.php
mbonneau: This has been moved to #115

davidwdan:
Closing this PR.  All other work will be individual PRs against the 2.x branch.

PR #117, opened 2017-01-07
Added more tests

davidwdan:
Removed some unused code
Misc cleanup

### DIFF
--- a/src/Disposable/BinaryDisposable.php
+++ b/src/Disposable/BinaryDisposable.php
@@ -30,9 +30,6 @@ class BinaryDisposable implements DisposableInterface
         $this->second = $second;
     }
 
-    /**
-     *
-     */
     public function dispose()
     {
         if ($this->isDisposed) {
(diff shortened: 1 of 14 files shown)


src/Disposable/BinaryDisposable.php
```diff
@@ -41,24 +38,12 @@ public function dispose()
...
-        }
-
-        $old2         = $this->second;
-        $this->second = null;
-        if ($old2) {
-            $old2->dispose();
-        }
+        $this->first->dispose();
 
+        $this->second->dispose();
```
mbonneau, line 42: We should probably still give up reference to the disposables here right?

src/Notification/OnNextNotification.php
```diff
@@ -26,19 +26,19 @@ protected function doAccept($onNext, $onError, $onCompleted)
...
 
-    public function __toString()
+    public function __toString(): string
     {
         return 'OnNext(' . json_encode($this->value) . ')';
     }
-    
-    public function equals($other)
+
+    public function equals($other): string
```
mbonneau: bool?

PR #119, opened 2017-01-09
Create specific exception for `timeout`

### LINKED ISSUE #113: Create specific exception for `timeout`
Create a specific exception so a timeout exception can be detected downstream as a timeout.

mbonneau:
This PR creates the `TimeoutException` so that a timeout can be identified downstream.

Fixes #113

### DIFF
--- a/src/TimeoutException.php
+++ b/src/TimeoutException.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Rx;
+
+class TimeoutException extends \Exception
+{
+}
--- a/src/TimeoutException.php
+++ b/src/TimeoutException.php
@@ -1,7 +1,0 @@
-<?php
-
-namespace Rx;
-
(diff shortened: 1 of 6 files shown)


src/TimeoutException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Rx;
+
+class TimeoutException extends \Exception
```
asm89: Do we want to have a base exception in the library while we're at it?

asm89:
lgtm 👍

mbonneau:
Base exception sounds good to me - I debated putting it in its own namespace also - @asm89 what do you think `Rx\TimeoutException` vs `Rx\Exception\TimeoutException`?

asm89:
Moving exceptions to their own namespace is a common pattern for libraries. Let's do it?

PR #125, opened 2017-01-11
Added isEmpty operator

martinsik:
This is RxJS 5 implementation of `isEmpty()` operator. Tests are from RxJS 5 as well.

I'm pushing this only to the `master` branch. Should I make another commit for `2.x` or wait until it's stable and then start adding more functionality?

### DIFF
--- a/demo/isEmpty/isEmpty-false.php
+++ b/demo/isEmpty/isEmpty-false.php
@@ -5,4 +5,4 @@ require_once __DIR__ . '/../bootstrap.php';
 $source = \Rx\Observable::just(1)
     ->isEmpty();
 
-$subscription = $source->subscribe($stdoutObserver);
+$source->subscribe($stdoutObserver);
--- a/demo/isEmpty/isEmpty-false.php
+++ b/demo/isEmpty/isEmpty-false.php
@@ -1,0 +1,8 @@
+<?php
+
+require_once __DIR__ . '/../bootstrap.php';
+
+$source = \Rx\Observable::just(1)
+    ->isEmpty();
+
+$subscription = $source->subscribe($stdoutObserver);
--- a/demo/isEmpty/isEmpty.php
+++ b/demo/isEmpty/isEmpty.php
@@ -1,0 +1,8 @@
+<?php
+
+require_once __DIR__ . '/../bootstrap.php';
+
+$source = \Rx\Observable::emptyObservable()
+    ->isEmpty();
+
+$subscription = $source->subscribe($stdoutObserver);
--- a/demo/isEmpty/isEmpty.php
+++ b/demo/isEmpty/isEmpty.php
@@ -5,4 +5,4 @@ require_once __DIR__ . '/../bootstrap.php';
 $source = \Rx\Observable::emptyObservable()
     ->isEmpty();
 
-$subscription = $source->subscribe($stdoutObserver);
+$source->subscribe($stdoutObserver);
--- a/lib/Rx/Observable.php
+++ b/lib/Rx/Observable.php
@@ -1859,7 +1859,7 @@ class Observable implements ObservableInterface
      * @demo isEmpty/isEmpty.php
      * @demo isEmpty/isEmpty-false.php
      * @operator
-     * @reactivex isEmpty
+     * @reactivex contains
      */
     public function isEmpty()
     {
--- a/lib/Rx/Observable.php
+++ b/lib/Rx/Observable.php
@@ -33,6 +33,7 @@ use Rx\Operator\DistinctOperator;
 use Rx\Operator\DistinctUntilChangedOperator;
 use Rx\Operator\DoOnEachOperator;
 use Rx\Operator\GroupByUntilOperator;
+use Rx\Operator\IsEmptyOperator;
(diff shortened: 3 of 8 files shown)


demo/isEmpty/isEmpty-false.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+require_once __DIR__ . '/../bootstrap.php';
+
+$source = \Rx\Observable::just(1)
+    ->isEmpty();
+
+$subscription = $source->subscribe($stdoutObserver);
```
davidwdan: You don't need `$subscription =` here

demo/isEmpty/isEmpty.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+require_once __DIR__ . '/../bootstrap.php';
+
+$source = \Rx\Observable::emptyObservable()
+    ->isEmpty();
+
+$subscription = $source->subscribe($stdoutObserver);
```
davidwdan: Same as above

lib/Rx/Operator/IsEmptyOperator.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Rx\Operator;
+
+use Rx\Observer\CallbackObserver;
```
davidwdan: The following use statements are not used:
```PHP
use Rx\Operator\OperatorInterface;
use Rx\Scheduler\ImmediateScheduler;
```

lib/Rx/Operator/IsEmptyOperator.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * @param \Rx\ObservableInterface $observable
+     * @param \Rx\ObserverInterface $observer
+     * @param \Rx\SchedulerInterface $scheduler
+     * @return \Rx\DisposableInterface
+     */
+    public function __invoke(ObservableInterface $observable, ObserverInterface $observer, SchedulerInterface $scheduler = null)
+    {
+        return $observable->subscribeCallback(
```
davidwdan: Should probably use `subscribe` here since `subscribeCallback` is not on the observable interface

lib/Rx/Observable.php
```diff
@@ -1849,4 +1850,21 @@ public function throttle($throttleDuration, $scheduler = null)
...
+
+    /**
+     * If the source Observable is empty it returns an Observable that emits true, otherwise it emits false.
+     *
+     * @return AnonymousObservable
+     *
+     * @demo isEmpty/isEmpty.php
+     * @demo isEmpty/isEmpty-false.php
+     * @operator
+     * @reactivex isEmpty
```
davidwdan: isEmpty is a variant of [contains](http://reactivex.io/documentation/operators/contains.html), so that should be 
`* @reactivex contains`

test/Rx/Functional/Operator/IsEmptyTest.php
```diff
@@ -0,0 +1,170 @@
...
+        }, 500);
+
+        $this->assertMessages([], $results->getMessages());
+
+        $this->assertSubscriptions([
+            subscribe(200, 500),
+        ], $xs->getSubscriptions());
+    }
+
+}
```
davidwdan, line 194: Can you add a test that never completes?

test/Rx/Functional/Operator/IsEmptyTest.php
```diff
@@ -0,0 +1,170 @@
...
+class IsEmptyTest extends FunctionalTestCase
+{
+
+    /**
+     * @test
+     */
+    public function should_return_true_if_source_is_empty()
+    {
+        $xs = $this->createHotObservable([
+            onCompleted(300)
```
davidwdan, line 16: Since this is a hot observable, can you add an onNext before 200, so we can verify that nothing emits before it's suppose to?

davidwdan (commented):
Looks good.  Just a couple of minor changes.

test/Rx/Functional/Operator/IsEmptyTest.php
martinsik: This should be already covered by [should_not_completes_if_source_never_emits ](https://github.com/ReactiveX/RxPHP/pull/125/files/d78d3a3d84bce819ac0e936f8b73f8a1bc4b7574#diff-b99fb56076d5d32fbf2472d605dbf506R85) test.

test/Rx/Functional/Operator/IsEmptyTest.php
martinsik: A similar situation is in [should_return_false_if_source_emits_element](https://github.com/ReactiveX/RxPHP/pull/125/files/d78d3a3d84bce819ac0e936f8b73f8a1bc4b7574#diff-b99fb56076d5d32fbf2472d605dbf506R36) but I added a separate test for this.

martinsik:
I elaborated everything you mentioned.  
If you don't mind I also updated `demo/bootstrap.php` to properly output `0` or `1` for booleans.

davidwdan:
Merged.  Can you open a PR against 2.x?

PR #129, opened 2017-01-23
The finally() operator

martinsik:
This is the `finally()` operator based on its RxJS 5 implementation. I called in `finallyCall()` here because `finally` is a keyword but in RxPHP 2 I'll rename it to just `finally()`.


felixfbecker:
In Promise implementations like reactphp or sabre this is typically called `always`

davidwdan:
@martinsik This is a `do` variant, so it might make sense to either call this `finallyDo` or `doOnTerminate`, which are names used by other Rx projects: http://reactivex.io/documentation/operators/do.html

If we want to stick to RxJS naming, `finallyAction` is an alias in RxJS 4.

@felixfbecker `always` doesn't make sense in this case, since we only want to call this when the observable sequence terminates, not each time it emits, which is what the name `always` implies.

davidwdan:
@martinsik it looks like there is a line break issue with the error demo with hhvm: https://travis-ci.org/ReactiveX/RxPHP/jobs/194568201#L341-L350

felixfbecker:
`doFinally` sounds good

davidwdan:
Just to clarify, I'm only suggesting the name change for RxPHP v1x.  I think that `finallyDo` makes the most sense.  @mbonneau?

felixfbecker:
Aren't all other `do` methods prefixed with `do` and not suffixed?

mbonneau:
I would go with `finallyDo`. I see no need to add a variant that doesn't already exist in any other rx project.

martinsik:
I think @felixfbecker has a good point. Isn't this going to be confusing when we'll have `doOnEach`, `doOnNext`, `doOnError`, `doOnCompleted` and then `finallyDo`?

mbonneau:
Just to move things along - I change my vote to `doFinally`

martinsik:
Ok, so I'll call it `doFinally` in RxPHP 1 and just `finally` in RxPHP 2.

PR #130, opened 2017-01-23
Added docblocks

martinsik:
I went through some classes and tried to add docblocks where missing. I checked the correct syntax for [`@param` and it always has to have proper type specified](https://www.phpdoc.org/docs/latest/references/phpdoc/tags/param.html#description). I've also seen `@override` which doesn't seem to be supported by phpdoc (isn't this from Java?).

This mostly applies to 2.x branch as well.

### DIFF
--- a/lib/Rx/Disposable/BinaryDisposable.php
+++ b/lib/Rx/Disposable/BinaryDisposable.php
@@ -10,10 +10,10 @@ use Rx\DisposableInterface;
  */
 class BinaryDisposable implements DisposableInterface
 {
-    /** @var \Rx\DisposableInterface */
+    /** @var DisposableInterface */
     private $first;
 
-    /** @var \Rx\DisposableInterface */
+    /** @var DisposableInterface */
     private $second;
 
     /** @var bool */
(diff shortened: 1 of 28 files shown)


davidwdan:
@martinsik I actually removed a lot of the doc blocks in 2.x because I think that they're unnecessary where we have clear typing.  I'm also not a big fan of `@inheritdoc`.

davidwdan:
If anything, I would like to update the doc blocks on the operators in `Observable` to include more info for things like callables.  

For example in the `create` doc block: 

```* @param callable(ObserverInterface $observer): DisposableInterface $subscribeAction```

martinsik:
Ok, do you want to also remove `@var` doc blocks? I think `@var` and `@param` are actually useful.  
The updated syntax for callable is still just a proposal I think phpDocumentor/fig-standards#19 (or is it already supported by IDEs?).

davidwdan:
`@var` doc blocks are okay.  Just make a judgement call on where it's not clear from the types in the constructor.

I'm not sure if IDE support it or not, but it does make sense from a documentation perspective. The async interop is also using that notation: https://github.com/async-interop/event-loop/blob/master/src/Loop.php#L142 .

martinsik:
So lets close this PR and I'll eventually open a new one.

PR #131, opened 2017-02-02
compose() operator

mamciek:
Please add thru() operator inspired by mostjs: https://github.com/cujojs/most/blob/master/docs/api.md#thru

```php
$process = function ($observable) {
    return $observable
        ->filter(function ($val) { return $val % 2 == 0; })
        ->map(function ($val) { return $val * 2; });
};
```

it allows to use

```php
$source = Observable::fromArray(range(1, 10))
       ->map(..)
       ->filter(..)
        ->thru($process)
        ->map(..);

```

instead of
```php
$source = $process(
                    Observable::fromArray(range(1, 10))
                       ->map(..)
                       ->filter(..)
                  )->map(..);
```


martinsik:
@mamciek I think you could write the same with `concatMap`:
```
Observable::fromArray(range(1, 10))
    ->concatMap(function($value) use ($process) {
        return $process(Observable::of($value));
    });
```

davidwdan:
Is this similar to RxJava's `compose` operator? I think that this operator can be useful for cleaning up code and is worth adding.

@mamciek can you add some tests?

mamciek:
I will add some tests. I wasn't aware of `compose` operator from RxJava. I only knew about `thru` in mostjs. I personaly think that `thru` is much better name, but maybe I should rename to `compose` in order to be more "compatible" with Rx family?

ddimaria:
+1 for compose as it has explicit meaning to FP programmers

asm89:
+1 for being more compatible with naming in the Rx family

PR #136, opened 2017-02-08
Custom operator in nested namespace

mamciek:
Allow customer operators in namespace like: \Vendor\Sub\Sub\Rx\Operator

### DIFF
--- a/lib/Rx/Observable.php
+++ b/lib/Rx/Observable.php
@@ -751,16 +751,21 @@ class Observable implements ObservableInterface
      * @param $name
      * @param $arguments
      * @return Observable
-     * 
+     *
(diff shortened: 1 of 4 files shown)


mamciek:
Added test.

mamciek:
Does anybody knows why Travis HHVM build does not pass? It fails with different demo/test.php every time.

mbonneau:
The HHVM fails intermittently on the 1.x (current master) branch. This is due to timing issues.

I have restarted the build.

davidwdan:
Maybe we should move HHVM to allow failures

mamciek:
@davidwdan I have added test. Is it ready to merge in your opinion?

PR #143, opened 2017-02-13
Allow plucking of multiple items at once

mbonneau:
This PR allows you to use pluck to pluck multiple levels deep:

```php
$x = [
    [
        [
            (object)[
                "something" => "Hello"
            ]
        ]
    ]
];

Rx\Observable::fromArray($x)->pluck(0, 0, 'something')->subscribe(function ($x) {
    echo $x;
});
```


davidwdan:
Should we back port this to v1?

PR #133, opened 2017-02-04
Marble testing

mbonneau:
This PR is a preliminary implementation of marble testing as requested at SunshinePHP at the talk by @luijar see: https://github.com/ReactiveX/rxjs/blob/master/doc/writing-marble-tests.md

Example:

```php
    public function testMapAddOne()
    {
        $cold     = '--1--2--|';
        $expected = '--2--3--|';

        $results = $this->scheduler->startWithCreate(function () use ($cold) {
            return $this->createCold($cold)->map(function ($x) { return $x + 1; });
        });

        $this->assertEquals($expected, $this->convertMessagesToMarbles($results->getMessages()));
    }
```

PR #144, opened 2017-02-15
Testing higher-order Observables

martinsik:
I was implementing the `window()` operator and I realized I've encountered a situation that wasn't tested before. That's higher-order Observables and in particular operators that emit Observables where I need to test that these Observables emit correct items themselves.

This PR lets me write tests for `window()` operator like the following:

```
$xs = $this->createHotObservable([
    onNext(225, 1),
    onNext(250, 2),
    onNext(300, 3),
    ...
]);

$results = $this->scheduler->startWithCreate(function() use ($xs, $window) {
    return $xs->window($window);
});

$window = $this->createHotObservable([
    onNext(325, 1),
   ...
]);

$this->assertMessages([
    onNext(200, $this->createColdObservable([
        onNext(25, 1),
        onNext(50, 2),
        onNext(100, 3),
        onCompleted(125),
    ])),
    ...
```

https://github.com/martinsik/RxPHP/blob/window-operator/test/Rx/Functional/Operator/WindowTest.php

The most important part is this:
```
onNext(200, $this->createColdObservable([
    onNext(25, 1),
    onNext(50, 2),
    onNext(100, 3),
```

There's a new `OnNextObservableNotification` that automatically subscribes to an Observable and all items are recorded with timestamps relative to its parent. Then these automatic subscriptions are ignored by the `ColdObservable`. I've actually spent quite some time scratching my head around how to implement this in the most comfortable and easy to use way. Ideally if I didn't need to think about it at all and use Observables just like any other values.

If I didn't miss anything this should allow fluent porting from RxJS 5 marble tests ([eg. window-spec.ts](https://github.com/ReactiveX/rxjs/blob/master/spec/operators/window-spec.ts)) where they treat Observables like values. For example this should work in RxPHP as well:

```
const expected =     'x------------y------------z------------|    ';
const x = cold(      '---a---b---c-|                              ');
const y = cold(                   '--d---e---f--|                 ');
const z = cold(                                '-g---h---i---|    ');
const expectedValues = { x: x, y: y, z: z };
```

Using this with ReactiveX/RxPHP#133 shouldn't be any problem.

Also, I didn't need to modify any existing tests including those that use higher-order Observables as input such as [`MergeAllTest.php`](https://github.com/ReactiveX/RxPHP/blob/master/test/Rx/Functional/Operator/MergeAllTest.php).

### DIFF
--- a/lib/Rx/Notification/OnNextObservableNotification.php
+++ b/lib/Rx/Notification/OnNextObservableNotification.php
@@ -11,14 +11,14 @@ class OnNextObservableNotification extends OnNextNotification
     /** @var MockHigherOrderObserver */
     private $observer;
 
-    public function __construct($value, TestScheduler $scheduler = null, $startTime = 0)
+    public function __construct($value, TestScheduler $scheduler = null)
     {
         parent::__construct($value);
         if (!$scheduler) {
             $scheduler = new TestScheduler();
         }
 
-        $this->observer = new MockHigherOrderObserver($scheduler, $startTime);
+        $this->observer = new MockHigherOrderObserver($scheduler, $scheduler->getClock());
 
         /** @var Observable $value */
         $value->subscribe($this->observer, $scheduler);
--- a/lib/Rx/Notification/OnNextObservableNotification.php
+++ b/lib/Rx/Notification/OnNextObservableNotification.php
@@ -3,13 +3,12 @@
 namespace Rx\Notification;
 
 use Rx\Observable;
-use Rx\SchedulerInterface;
-use Rx\Testing\MockObserver;
+use Rx\Testing\MockHigherOrderObserver;
 use Rx\Testing\TestScheduler;
 
 class OnNextObservableNotification extends OnNextNotification
 {
-    /** @var MockObserver */
+    /** @var MockHigherOrderObserver */
     private $observer;
 
     public function __construct($value, TestScheduler $scheduler = null, $startTime = 0)
@@ -19,7 +18,7 @@ class OnNextObservableNotification extends OnNextNotification
             $scheduler = new TestScheduler();
         }
 
-        $this->observer = new MockObserver($scheduler, $startTime);
+        $this->observer = new MockHigherOrderObserver($scheduler, $startTime);
 
         /** @var Observable $value */
         $value->subscribe($this->observer, $scheduler);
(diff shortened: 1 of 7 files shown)


martinsik:
As a proof of concept I cherry picked commits from #133 and my `window()` operator and made marble tests for `window()` operator just to see whether it really works the way I think it should.

https://github.com/martinsik/RxPHP/blob/marble-tests-window-operator/test/Rx/Functional/Operator/WindowTest.php

I had to [make a few small changes to `FunctionalTestCase` class](https://github.com/martinsik/RxPHP/commit/002671800236c56d872a68dc635a5836c4f39448#diff-5d025602e32be259357e6c4e45dd35b2) in the way it handles marbles and I also added parsing subscriptions. I'll make appropriate comments into #133.

For example this test is based on [`window-spec.ts`](https://github.com/ReactiveX/rxjs/blob/master/spec/operators/window-spec.ts) and is passing:

```
/**
 * @test
 */
public function should_emit_windows_that_close_and_reopen()
{
    $source = $this->createHot( '---a---b---c---d---e---f---g---h---i---|');
    $sourceSubs =               '^                                      !';
    $window = $this->createHot( '-------------w------------w------------|');
    $windowSub =                '^                                      !';
    $expected =                 'x------------y------------z------------|';
    $x      = $this->createCold('---a---b---c-|                          ');
    $y      = $this->createCold(             '--d---e---f--|             ');
    $z      = $this->createCold(                          '-g---h---i---|');
    $expectedValues = [ 'x' => $x, 'y' => $y, 'z' => $z ];

    $results = $this->scheduler->startWithCreate(function() use ($source, $window) {
        return $source->window($window);
    });

    $expectedMessages = $this->convertMarblesToMessages($expected, $expectedValues, null, TestScheduler::SUBSCRIBED);
    $this->assertMessages($expectedMessages, $results->getMessages());

    $this->assertSubscriptions($sourceSubs, $source->getSubscriptions(), TestScheduler::SUBSCRIBED);
    $this->assertSubscriptions($windowSub, $window->getSubscriptions(), TestScheduler::SUBSCRIBED);
}
```

Some things are a bit clumsy. I had to manually tell when the subscription happens with `TestScheduler::SUBSCRIBED` because that's predefined by `startWithCreate()`. Otherwise it thinks it subscribes at `0`.
Using `startWithTiming($create, 0, 0)` is probably not a good options because then the `TestScheduler` count `+1` automatically. I guess this is solvable somehow better that hardcoding `TestScheduler::SUBSCRIBED` like I did.

davidwdan (approved):

PR #150, opened 2017-02-27
[WIP] Rework Scheduler

mbonneau:
This PR returns the `EventLoopScheduler` to the v1 style construction.

This also adds `AsyncSchedulerInterface` for operators and observables that require async schedulers.

The README.md still needs to be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,22 +46,13 @@ $ php demo/interval/interval.php
 
 Have fun running the demos in `/demo`.
 
-note:  The demos are automatically run within `Loop::execute`.  When using RxPHP within your own project, you'll need to install a [loop implementation](https://packagist.org/providers/async-interop/event-loop-implementation).  
+note:  When running the demos, the scheduler is automatically bootstrapped.  When using RxPHP within your own project, you'll need to set the default scheduler; 
 
 ## Installation
-1. Install one [async-interop event loop](https://packagist.org/providers/async-interop/event-loop-implementation) implementation.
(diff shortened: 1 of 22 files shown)


martinsik:
I'm just curious what happened to `Loop` and `async-interop`. I read the discussions in `async-interop/event-loop` repo and it seems the API is likely to change and it'll take some for the first stable release. Is this the reason we're going back to `React\EventLoop`?

mbonneau:
@martinsik This change is being made because version 2 of RxPHP is taking too long to release with the dependency on the interop loop as the only blocking issue.

When first developing version 2 of RxPHP, it appeared that work on the Async-Interop Loop library had settled a bit and we had hoped that a stable version was close (see: https://github.com/async-interop/event-loop/issues/119). We chose to try and push things forward with adoption of the interop loop standard. Since then, there have been more issues raised with the interop loop and consensus is taking a bit longer than we would like for inclusion into this next release of RxPHP.

Inclusion of the interop loop would have allowed users to remove a small bit of boiler plate code. With this PR, construction of the actual scheduler will be the same as it was in v1 and then you can set the default scheduler in the `Scheduler` static. RxPHP is loop agnostic (react is a dev-dependency for testing).

davidwdan:
I'm going to merge this because the async interop loop project is all but dead.  If and when there is a loop standard we can introduce new interop loop scheduler under 2.x and then bump the version to v3 or what ever the next major is, when we're reading to make it the default.

PR #153, opened 2017-03-10
[WIP] Higher-Order Marble tests

davidwdan:
I cherry picked the higher-order test code from v1 so we can start working on getting higher-order marble tests working.

Some tests are failing.  
@martinsik if you get a chance, can you look at the failing tests?


martinsik:
@davidwdan Can you give me write permissions to your `marbleTest` branch?

davidwdan:
@martinsik you have permission now.  I think the issue is that in some circumstances there are multiple test schedulers.  If I hack it and pass the test scheduler through the `onNext` helper function, it works.

src/Notification/OnNextObservableNotification.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Rx\Notification;
+
+use Rx\Observable;
+use Rx\Testing\ColdObservable;
+use Rx\Testing\HotObservable;
+use Rx\Testing\MockHigherOrderObserver;
+use Rx\Testing\TestScheduler;
+
+class OnNextObservableNotification extends OnNextNotification
```
asm89: nit: should this live in `Rx\Testing` instead? It depends on the `TestScheduler`?

src/Testing/ColdObservable.php
```diff
@@ -26,13 +26,16 @@ public function __construct(TestScheduler $scheduler, array $messages = [])
...
-        $this->subscriptions[] = new Subscription($this->scheduler->getClock());
-        $index                 = count($this->subscriptions) - 1;
-
         $currentObservable = $this;
         $disposable        = new CompositeDisposable();
         $scheduler         = $this->scheduler;
         $isDisposed        = false;
+        $index             = null;
+
+        if (!($observer instanceof MockHigherOrderObserver)) {
```
asm89: Why do we need all these `instanceof` checks? Looks wrong.

asm89 (changes requested):
I'm probably missing something, but I'm not a fan of the `instanceof` checks scattered all over the code. Is there no other way to achieve this?

davidwdan:
Inner observables are now materialized when being converted to a string.  This simplifies things and gets rid of the need to check the instance of the observers.

The test `automatic_mock_observer_doesnt_create_loggable_subscription` is failing, but I not sure if it's still needed or when those inner subscriptions should start and stop.

test/helper-functions.php
```diff
@@ -2,28 +2,51 @@
...
     return new Subscription($start, $end);
 }
 
-function RxIdentity($x) {
+function RxIdentity($x)
+{
     return $x;
 }
+
+function materializeObservable(Observable $observable): array
```
asm89, line 39: `OnNextNotification` does not live in `testing`, but is now depending on this testing helpers file.

asm89: Could the dependencies be inversed? E.g. using `->accept()` on the notifications from the testing code?

martinsik:
@davidwdan That test is irrelevant because now the `MockObserver` automatically subscribes only when calling `OnNextNotification:: __toString()`.

It was previously used in situations where we don't want to log subscriptions. For example in tests such as [MergeAllTest.php#L14](https://github.com/ReactiveX/RxPHP/blob/master/test/Rx/Functional/Operator/MergeAllTest.php#L14) where it's not a problem anymore because there are no `__toString()` calls on `$xs` or `$ys`.

test/helper-functions.php
davidwdan: @asm89 ACK I'll refactor it in a bit.  Aren't we planning to move some of the test helpers out of the testing namespace?

davidwdan:
I added a `groupBy` marble test that is currently failing.  We may have to change where the inner observable is materialized

PR #155, opened 2017-03-14
Revert back to using React Promises

davidwdan:
There doesn't seem to be much of a consensus around the interop promise spec and amp will be supporting [React promises](https://github.com/amphp/amp/pull/76), so I'm going to revert promises back to just supporting react.


src/React/Promise.php
```diff
@@ -0,0 +1,99 @@
...
+{
+    /**
+     * @param mixed $value
+     * @return ReactPromise A promise resolved to $value
+     */
+    public static function resolved($value): ReactPromise
+    {
+        $d = new Deferred();
+        $d->resolve($value);
+        return $d->promise();
```
kelunik, line 25: This can just be `new FulfilledPromise($value)` in React.

PR #156, opened 2017-03-15
Updated README in prep for v2 release

davidwdan:


docs/build-docs.php
```diff
@@ -34,7 +34,12 @@ function run_lint() {
 }
 
 function run_reactivex() {
-    $docs = load_all_docs();
+    $allDocs = load_all_docs();
+
+    $docs = array_filter($allDocs, function ($doc) {
+        echo "d ", $doc->isDeprecated;
```
mbonneau: Is this echo needed?

davidwdan: Nope... fixed

PR #158, opened 2017-03-15
Fix order of subscriptions in the takeUntil() operator

martinsik:
This switches the order of subscriptions in the `takeUntil()` operator. This is relevant only when using the `ImmediateScheduler`.

```
$source = Observable::range(1, 10);
$notification = new Subject();

$source->takeUntil($notification)
    ->subscribe(new CallbackObserver(function($value) use ($notification) {
        echo "$value\n";
        if ($value === 5) {
            $notification->onNext(true);
        }
    }));
```

This will print numbers `1 - 10` even though I called `$notification->onNext(true)` after the first five items because the `takeUntil()` operator first [subscribes to its source](https://github.com/ReactiveX/RxPHP/blob/master/src/Operator/TakeUntilOperator.php#L26) which starts emitting values immediatelly before subscribing to the notifier.

### DIFF
--- a/lib/Rx/Operator/TakeUntilOperator.php
+++ b/lib/Rx/Operator/TakeUntilOperator.php
@@ -28,14 +28,14 @@ class TakeUntilOperator implements OperatorInterface
     {
 
         return new CompositeDisposable([
-            $observable->subscribe($observer, $scheduler),
             $this->other->subscribe(
                 new CallbackObserver(
                     [$observer, "onCompleted"],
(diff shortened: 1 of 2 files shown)


martinsik:
@davidwdan I guess I should send PR agains the `2.x` as well...

davidwdan:
@martinsik that would be great.  Would you also be interested in back porting #167 and #168 to v1?

martinsik:
@davidwdan I'll have a look at all of them.

PR #167, opened 2017-04-04
Dispose of event loop timers and never create more than one

### LINKED ISSUE #165: Items scheduled outside if scheduler can cause memory leak with EventLoopScheduler
In the `EventLoopScheduler`, if items are scheduled from non-scheduled events (like I/O), additional timers may be created and not properly disposed of.

mbonneau:
Fixes #165 .

Fixes `EventLoopScheduler` to only ever schedule 1 timer with the underlying loop. Also disposes of unneeded timers.

This has a small API change with the EventLoopScheduler now requiring the delay callback (if a callback is used) to return a `DisposableInterface` that can cancel the actual timer.

### DIFF
--- a/src/Scheduler/EventLoopScheduler.php
+++ b/src/Scheduler/EventLoopScheduler.php
@@ -5,6 +5,8 @@ declare(strict_types = 1);
 namespace Rx\Scheduler;
 
 use React\EventLoop\LoopInterface;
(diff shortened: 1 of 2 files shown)


davidwdan:
The Amp docs need to be updated or removed until Amp v2 is released

davidwdan:
This fix probably needs to be back ported to v1.

PR #169, opened 2017-04-05
Backport #167 and #168 to 1.x

martinsik:
#167 #168


martinsik:
I'm not sure why that one test is failing only for PHP 5.6.  
Isn't this some random timing issue similar to what we had with running demo scripts?

davidwdan:
That does look like a timing issue.

PR #175, opened 2017-05-19
Fixed memory leak in GroupByUntil and refactored GroupedObservable

davidwdan:


src/Disposable/RefCountDisposable.php
```diff
@@ -42,25 +42,23 @@ public function getDisposable()
...
-    private function createInnerDisposable()
+    private function createInnerDisposable(): DisposableInterface
     {
-        $count = &$this->count;
-        $count++;
-        $innerDisposable      = &$this;
-        $isInnerDisposed      = false;
-        $underLyingDisposable = &$this->disposable;
+        $this->count++;
+        $innerDisposable = &$this;
```
mbonneau: Can we drop this and just use $this?

src/Disposable/RefCountDisposable.php
```diff
@@ -42,25 +42,23 @@ public function getDisposable()
...
+        $isInnerDisposed = false;
 
-        return new CallbackDisposable(function () use (&$count, &$innerDisposable, &$isInnerDisposed) {
+        return new CallbackDisposable(function () use (&$innerDisposable, &$isInnerDisposed) {
             if ($innerDisposable->isDisposed()) {
```
mbonneau: $this

src/Operator/GroupByUntilOperator.php
```diff
@@ -58,123 +51,108 @@ public function __construct(callable $keySelector, callable $elementSelector = n
...
+                            $observer->onError($e);
                         },
-                        function () use ($expire) {
-                            $expire();
-                        }
-                    );
-
-                    $subscription = $duration->take(1)->subscribe($callbackObserver);
+                        function () use ($groupDisposable, $md, $writer, $serializedKey) {
+                            if (isset($this->map[$serializedKey])) {
```
mbonneau: Would this ever be false?

PR #177, opened 2017-08-09
Proposal: Support Amp and React promises

trowski:
This PR adds support for Amp promises as well as decoupling RxPHP from React. Users of RxPHP may choose to install `react/promise`, `amphp/amp`, or both. Alternatively, if users do not need to use observables with promises, there is no need to install either.

While it is currently possible to use RxPHP with Amp, there are a few annoyances. Specifically, `Observable::fromPromise()` only accepts React promises, making it a hassle to use with Amp. This PR simplifies this usage without complicating it for React users.

Note that this is a WIP, primarily because more tests should be added using Amp promises, but I wanted to get some feedback before putting in more effort.


davidwdan:
Thanks for this PR.  This can be accomplished without making any changes to RxPHP.  I'd much rather see you add a helper function to Amp that converts an Amp promise to an observable.

I don't see that big of a different in clunkiness between `Observable::fromPromise()` and `Amp::toObservable()` or whatever you end up calling it.  

Another option would be to create a separate project that bootstraps the default scheduler with the Amp loop and provides the `toObservable` function, so you'd only have to include that one project and everything would work.

Example:

```PHP
use Amp\Artax\Response;
use Amp\Loop;
use Rx\Disposable\CallbackDisposable;
use Rx\Observable;
use Rx\React\RejectedPromiseException;
use Rx\Scheduler;
use Rx\Scheduler\EventLoopScheduler;
use Rx\Subject\AsyncSubject;

//You only need to set the default scheduler once.  This can be moved to a separate project that bootstraps this for the user.
Scheduler::setDefaultFactory(function () {
    return new EventLoopScheduler(function (int $ms, callable $callable) {
        $timer = Loop::delay($ms, $callable);
        return new CallbackDisposable(function () use ($timer) {
            Loop::cancel($timer);
        });
    });
});

Loop::run(function () {
    echo yield Observable::interval(1000)
        ->flatMap(function ($i) {
            return toObservable(
                (new Amp\Artax\DefaultClient)->request("https://httpbin.org/get?test_arg={$i}")
            );
        })
        ->flatMap(function (Response $response) {
            return toObservable($response->getBody());
        })
        ->take(1) // or reduce it down to a single value
        ->toPromise();
});

function toObservable(\Amp\Promise $promise)
{
    $subject = new AsyncSubject();

    $promise->onResolve(function ($error, $value) use ($subject) {
        if ($error) {
            $error = $error instanceof \Exception ? $error : new RejectedPromiseException($error);
            $subject->onError($error);
            return;
        }

        $subject->onNext($value);
        $subject->onCompleted();
    });

    return $subject->asObservable();
}

```

If Amp picks up in popularity, we can revisit decoupling React promises and including support for Amp in the next major release.

PR #185, opened 2017-09-05
Added singleInstance operator

davidwdan:
Current tests match RxJS.  We can add more tests later if needed.


davidwdan:
The failing test is unrelated to this PR and will be fixed in #186

PR #191, opened 2017-10-04
Dispose inner observable when promise is cancelled

### LINKED ISSUE #178: Promise::fromObservable should be cancelable
https://github.com/ReactiveX/RxPHP/blob/1838ede703bcb14eb67ed54cd61401f2ffff08c8/src/React/Promise.php#L49

Promises that are constructed here should be cancelable so that the user can dispose the underlying subscription.

davidwdan:
Fixes #178

### DIFF
--- a/test/Rx/Functional/Promise/ToPromiseTest.php
+++ b/test/Rx/Functional/Promise/ToPromiseTest.php
@@ -94,7 +94,14 @@ class ToPromiseTest extends FunctionalTestCase
      */
     public function promise_cancel()
     {
-        $promise = Observable::timer(1000)->mapTo(42)->toPromise();
+        $disposed = false;
+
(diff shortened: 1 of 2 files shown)


test/Rx/Functional/Promise/ToPromiseTest.php
```diff
@@ -87,4 +87,23 @@ function (Exception $ex) use (&$error) {
 
         $this->assertEquals(new Exception('some error'), $error);
     }
+
+    /**
+     * @test
+     *
+     */
+    public function promise_cancel()
```
mbonneau, line 95: Can we prove that the upstream is disposed of also? This just proves that the promise didn't resolve.

davidwdan: Done

PR #208, opened 2020-12-11
Allow PHP 8 and Update PHPUnit

mbonneau:


.travis.yml
```diff
@@ -50,4 +15,4 @@ script:
   - php docs/build-docs.php lint
 
 after_success:
-  - travis_retry php vendor/bin/coveralls -v
+  - travis_retry php vendor/bin/coveralls -v
```
asm89: nit: revert?

docs/doc-loader.php
```diff
@@ -77,7 +77,8 @@ public static function fromPath($path) {
         $outputPath = __DIR__ . '/../demo/' . $path . '.expect';
 
         assert(file_exists($codePath), "code does not exist $codePath");
-        assert(file_exists($outputPath), "output does not exist $outputPath");
+        // allow demos without expect files - we will just get a warning later
+        // assert(file_exists($outputPath), "output does not exist $outputPath");
```
asm89, line 81: Why does this need to be removed? Perhaps we can skip and print a warning for now if the output path does not exist?

src/Scheduler/VirtualTimeScheduler.php
```diff
@@ -21,7 +21,7 @@ class VirtualTimeScheduler implements AsyncSchedulerInterface
      * @param integer $initialClock Initial value for the clock.
      * @param callable $comparer Comparer to determine causality of events based on absolute time.
      */
-    public function __construct(int $initialClock = 0, callable $comparer)
+    public function __construct(int $initialClock, callable $comparer)
```
asm89, line 24: Interesting. I guess this went unnoticed before.

asm89 (approved):
LGTM!

docs/doc-loader.php
mbonneau: In PHP 7 this generated a warning and the tests still passed. With PHP 8 this generates a fatal error and stops.

A warning is still generated later when `file_get_contents` is called, but the tests still work.

.travis.yml
mbonneau: I don't remember removing that. I will take a look.

PR #219, opened 2023-12-08
Check if key/property exists in array/object

### LINKED ISSUE #216: Pluck will error on null value
Pluck uses `isset` to check for the existence of a key. `isset` will not return `true` if the associated value is null which causes `pluck` to error.

This should use `array_key_exists` instead. (`property_exists` for objects)

Example:
```php
\Rx\Observable::of([null])
    ->pluck(0)
    ->subscribe(
        fn($x) => printf("Next: %s\n", json_encode($x)),
        fn(\Throwable $e) => printf("Error: %s\n", $e->getMessage()),
        fn() => printf("Completed.\n")
    );
```
Outputs:
```
Error: Unable to pluck "0"
```

The expected output would be:
```
Next: null
Completed.
```

---
Here is the object version which exhibits the same behavior:
```php
\Rx\Observable::of((object)['foo' => null])
    ->pluck('foo')
    ->subscribe(
        fn($x) => printf("Next: %s\n", json_encode($x)),
        fn(\Throwable $e) => printf("Error: %s\n", $e->getMessage()),
        fn() => printf("Completed.\n")
    );
```

mCassy:
- Check if key/property exists in array/object
- Fix indentation

Closes #216

### DIFF
--- a/src/Observable.php
+++ b/src/Observable.php
@@ -1,6 +1,6 @@
 <?php
 
-declare(strict_types = 1);
+declare(strict_types=1);
 
 namespace Rx;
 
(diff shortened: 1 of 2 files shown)


mbonneau:
Thank @mCassy - sorry it took so long to merge.

PR #225, opened 2025-06-21
PHP 8.4

makomweb:
- uses Rector to bring the code base up to being PHP 8.4 compliant
- fixes a bug with an optional default scheduler being null

### DIFF
--- a/src/Disposable/SingleAssignmentDisposable.php
+++ b/src/Disposable/SingleAssignmentDisposable.php
@@ -27,7 +27,7 @@ class SingleAssignmentDisposable implements DisposableInterface
         }
     }
 
-    public function setDisposable(DisposableInterface $disposable = null)
+    public function setDisposable(?DisposableInterface $disposable = null)
     {
         if ($this->current) {
             throw new RuntimeException('Disposable has already been assigned.');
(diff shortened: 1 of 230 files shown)


bartvanhoutte:
Can you double check the following please?

`PHP Deprecated:  Return type of Rx\Scheduler\InternalPriorityQueue::insert($item, $priority) should either be compatible with SplPriorityQueue::insert(mixed $value, mixed $priority): true, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice`

makomweb:
@bartvanhoutte What do I have to do in order to see this deprecation message?

daifma:
What is the status of this PR ? do you need a help ?

makomweb:
@daifma Yes, I have no idea what is missing to get this merged. All kinds of advice would be awesome.

src/Disposable/SingleAssignmentDisposable.php
```diff
@@ -27,7 +27,7 @@ public function dispose()
         }
     }
 
-    public function setDisposable(DisposableInterface $disposable = null)
+    public function setDisposable(?DisposableInterface $disposable = null)
```
mbonneau, line 30: Nullable parameter type is not supported in 7.0.0 - we should bump the composer.json php version up to 7.1 or higher.

PR #227, opened 2026-03-10
Introduce action that automates CI version matrix

WyriHaximus:


WyriHaximus:
When creating this PR, the assumption was that the PHP requirement in `composer.json` matches what the code supports. Should I update `composer.json` to 8.1+?

mbonneau:
@WyriHaximus Yes, I have not problem bumping up, especially since 8.1 is past end-of-life already.

WyriHaximus:
@mbonneau Done, did the removal of everything below 8.1 in its own commit

